PR #22, opened 2022-04-26
Add support for non-opaque geometry.

maoliver-amd:
This sets the opaque flag onto geometry in the spatial structure to improve traversal performance in the case that an objects material does not contain an alpha mask.
This enables quickly detecting alpha masked geometry during ray traversal.

The main changes are that the gltf loader will now not automatically convert images to 4 channel. This allows us to detect which images have an alpha channel in the source and use that to detect non-opaque geometry. The conversion to 4 channel is than moved to gltf.

The RaytracingPrimitiveBuild functions now take an additional parameter used to signal to the spatial structure whether that object contains opaque or non-opaque geometry by setting the corresponding flag (before this was unset which flags everything as non-opaque)

### DIFF
--- a/gfx.h
+++ b/gfx.h
@@ -204,7 +204,7 @@ GfxResult gfxRaytracingPrimitiveBuild(GfxContext context, GfxRaytracingPrimitive
 GfxResult gfxRaytracingPrimitiveSetTransform(GfxContext context, GfxRaytracingPrimitive raytracing_primitive, float const *row_major_4x4_transform);
 GfxResult gfxRaytracingPrimitiveSetInstanceID(GfxContext context, GfxRaytracingPrimitive raytracing_primitive, uint32_t instance_id);   // retrieved through `ray_query.CommittedInstanceID()`
 GfxResult gfxRaytracingPrimitiveSetInstanceMask(GfxContext context, GfxRaytracingPrimitive raytracing_primitive, uint8_t instance_mask);
-GfxResult gfxRaytracingPrimitiveUpdate(GfxContext context, GfxRaytracingPrimitive raytracing_primitive, uint32_t flags = 0);
+GfxResult gfxRaytracingPrimitiveUpdate(GfxContext context, GfxRaytracingPrimitive raytracing_primitive);
 
 //!
 //! Draw state manipulation.
@@ -2251,14 +2251,14 @@ public:
         return kGfxResult_NoError;
     }
 
-    GfxResult updateRaytracingPrimitive(GfxRaytracingPrimitive const &raytracing_primitive, uint32_t flags)
+    GfxResult updateRaytracingPrimitive(GfxRaytracingPrimitive const &raytracing_primitive)
     {
         if(dxr_device_ == nullptr)
             return kGfxResult_InvalidOperation; // avoid spamming console output
         if(!raytracing_primitive_handles_.has_handle(raytracing_primitive.handle))
             return GFX_SET_ERROR(kGfxResult_InvalidParameter, "Cannot update an invalid raytracing primitive object");
         RaytracingPrimitive &gfx_raytracing_primitive = raytracing_primitives_[raytracing_primitive];
-        return buildRaytracingPrimitive(raytracing_primitive, gfx_raytracing_primitive, true, flags);
+        return buildRaytracingPrimitive(raytracing_primitive, gfx_raytracing_primitive, true, 0);
     }
 
     GfxProgram createProgram(char const *file_name, char const *file_path, char const *shader_model)
@@ -7447,11 +7447,11 @@ GfxResult gfxRaytracingPrimitiveSetInstanceMask(GfxContext context, GfxRaytracin
     return gfx->setRaytracingPrimitiveInstanceMask(raytracing_primitive, instance_mask);
 }
 
-GfxResult gfxRaytracingPrimitiveUpdate(GfxContext context, GfxRaytracingPrimitive raytracing_primitive, uint32_t flags)
+GfxResult gfxRaytracingPrimitiveUpdate(GfxContext context, GfxRaytracingPrimitive raytracing_primitive)
 {
     GfxInternal *gfx = GfxInternal::GetGfx(context);
     if(!gfx) return kGfxResult_InvalidParameter;
-    return gfx->updateRaytracingPrimitive(raytracing_primitive, flags);
+    return gfx->updateRaytracingPrimitive(raytracing_primitive);
 }
 
 GfxDrawState::GfxDrawState()
--- a/gfx.h
+++ b/gfx.h
@@ -790,6 +790,7 @@ class GfxInternal
         uint32_t vertex_stride_ = 0;
         GfxBuffer vertex_buffer_ = {};
         GfxAccelerationStructure acceleration_structure_ = {};
+        uint32_t flags = 0;
     };
     GfxArray<RaytracingPrimitive> raytracing_primitives_;
     GfxHandles raytracing_primitive_handles_;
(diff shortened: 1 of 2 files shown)


gboisse:
Hey Matthew,

So a couple of feedback:

I don't think we should rely on the channel count at all for detecting alpha textures; albedo maps are often 4-channel in my experience, but with alpha set to 1 everywhere.
I'd suggest we rather check if there's any non-opaque texel upon importing the texture and add some flags to the `GfxImage` object to communicate this information; relying on channel count seems fragile to me.

Second feedback is that I'd rather have flags instead of that `bool non_opaque` boolean.
We could also have the flags default to `0` (e.g., `uint32_t build_flags = 0`) to preserve the previously optional parameters 🙂

maoliver-amd:
> I don't think we should rely on the channel count at all for detecting alpha textures; albedo maps are often 4-channel in my experience, but with alpha set to `1` everywhere. I'd suggest we rather check if there's any non-opaque texel upon importing the texture and add some flags to the `GfxImage` object (rather than a `bool`) to communicate this information; relying on the channel count seems fragile to me.

This was mainly just a performance thing to avoid having to check the alpha on every texture. Since with Capsaicin we control all the test scenes we can control the textures to only have alpha channels were relevant. However, if you want gfx to be more general then Ill keep the current changes so we can fast skip 3 channel textures but add an extra check for pre-existing 4 channel where during copy we check that the alpha layer is actually valid.

> Second feedback is that I'd rather have flags instead of that `bool non_opaque` boolean on the BVH build functions. We could also have the flags default to `0` (e.g., `uint32_t build_flags = 0`) to preserve the previously optional parameters 🙂

I wasnt 100% about the way this was done either.

maoliver-amd:
OK made the requested changes, let me know how that looks.

maoliver-amd:
So I also added delta light types to gfx so they can be loaded from gltf files (obj dont support them)
The gfxLight type is currently pretty much identical to the gltf light type with respect to what data it represents. So for instance it currently only supports a single r^2 falloff parameter. If you ever wanted to add fbx scene loading or something like that later then this can be updated to have 3 falloff values (old school constant, linear and quadratic values) unless you want that support now

gfx.h
```diff
@@ -191,15 +191,23 @@ GfxRaytracingPrimitive const *gfxAccelerationStructureGetRaytracingPrimitives(Gf
 
 class GfxRaytracingPrimitive { GFX_INTERNAL_NAMED_HANDLE(GfxRaytracingPrimitive); public: };
 
+enum GfxRaytracingPrimitiveFlags
+{
+    GfxRaytracingPrimitiveFlags_None = 0,
+    GfxRaytracingPrimitiveFlags_NonOpaque = 1,
+
+    GfxRaytracingPrimitiveFlags_Count
+};
```
gboisse, line 200: Code convention is to start the constant name with a `k` here.

Also, what about keeping the flags defined the same way than say the `GfxCreateContextFlag` enum for instance?
So something more like this:
```
enum GfxBuildRaytracingPrimitiveFlag
{
    kGfxBuildRaytracingPrimitiveFlag_Opaque = 1 << 0
};
```

Therefore, if you do not pass the `kGfxBuildRaytracingPrimitiveFlag_Opaque` bit flag, the BVH is considered non-opaque.
Will allow to add possibly more flags later on.

gfx.h
```diff
@@ -191,15 +191,23 @@ GfxRaytracingPrimitive const *gfxAccelerationStructureGetRaytracingPrimitives(Gf
...
+
 GfxRaytracingPrimitive gfxCreateRaytracingPrimitive(GfxContext context, GfxAccelerationStructure acceleration_structure);
 GfxResult gfxDestroyRaytracingPrimitive(GfxContext context, GfxRaytracingPrimitive raytracing_primitive);
 
-GfxResult gfxRaytracingPrimitiveBuild(GfxContext context, GfxRaytracingPrimitive raytracing_primitive, GfxBuffer vertex_buffer, uint32_t vertex_stride = 0);
-GfxResult gfxRaytracingPrimitiveBuild(GfxContext context, GfxRaytracingPrimitive raytracing_primitive, GfxBuffer index_buffer, GfxBuffer vertex_buffer, uint32_t vertex_stride = 0);
+GfxResult gfxRaytracingPrimitiveBuild(GfxContext context, GfxRaytracingPrimitive raytracing_primitive, GfxBuffer vertex_buffer, uint32_t vertex_stride = 0, GfxRaytracingPrimitiveFlags flags = GfxRaytracingPrimitiveFlags_NonOpaque);
```
gboisse, line 205: The type should be `uint32_t` since you can combine multiple flags possibly (although a single one exists atm 😛).

So either `uint32_t build_flags = 0` or `uint32_t build_flags = kGfxBuildRaytracingPrimitiveFlag_Opaque` depending on what we'd like the default behaviour to be.
More leaning towards `0` being the default atm.

gfx.h
```diff
@@ -191,15 +191,23 @@ GfxRaytracingPrimitive const *gfxAccelerationStructureGetRaytracingPrimitives(Gf
...
+GfxResult gfxRaytracingPrimitiveBuild(GfxContext context, GfxRaytracingPrimitive raytracing_primitive, GfxBuffer index_buffer, GfxBuffer vertex_buffer, uint32_t vertex_stride = 0, GfxRaytracingPrimitiveFlags flags = GfxRaytracingPrimitiveFlags_NonOpaque);
 GfxResult gfxRaytracingPrimitiveSetTransform(GfxContext context, GfxRaytracingPrimitive raytracing_primitive, float const *row_major_4x4_transform);
 GfxResult gfxRaytracingPrimitiveSetInstanceID(GfxContext context, GfxRaytracingPrimitive raytracing_primitive, uint32_t instance_id);   // retrieved through `ray_query.CommittedInstanceID()`
 GfxResult gfxRaytracingPrimitiveSetInstanceMask(GfxContext context, GfxRaytracingPrimitive raytracing_primitive, uint8_t instance_mask);
-GfxResult gfxRaytracingPrimitiveUpdate(GfxContext context, GfxRaytracingPrimitive raytracing_primitive);
+GfxResult gfxRaytracingPrimitiveUpdate(GfxContext context, GfxRaytracingPrimitive raytracing_primitive, GfxRaytracingPrimitiveFlags flags = GfxRaytracingPrimitiveFlags_NonOpaque);
```
gboisse, line 210: Do we really need to pass the flags again on update?
I would have thought having them on build should suffice; do you think it's beneficial to be able to change those on update?

gfx_scene.h
```diff
@@ -175,13 +175,22 @@ GfxMetadata const &gfxSceneGetCameraMetadata(GfxScene scene, uint64_t camera_han
 //! Image object.
 //!
 
+enum GfxImageFlags
+{
+    kGfxImageFlags_None = 0,
+    kGfxImageFlags_AlphaChannel = 1,
+
+    kGfxImageFlags_Count
+};
```
gboisse, line 184: Bit same feedback than the previous enum, I think I'd rather see:
```
enum GfxImageFlag
{
    kGfxImageFlag_HasAlpha = 1 << 0
};
```
With the `uint32_t` flags being initialized to `0`.
What do you think?

gboisse (changes requested):

gfx.h
maoliver-amd: Yep that makes sense. I had based the enum on `GfxCpuAccess` which is why it was the way it was.

gfx.h
maoliver-amd: Ive set this as default `0`. There are performance advantages to setting geometry as opaque which is why it could be useful as the default however then you have the potential issue of alpha masked geometry not being detected if not explicitly set to `0`. So either way the user should use the correct flag for best results (and a default '0' is less code)

gfx.h
maoliver-amd: The only reason thats there is because `updateRaytracingPrimitive` calls `buildRaytracingPrimitive` which requires the flag parameter be passed. Should the update fail then it creates a new one which requires that flag. So I wasnt sure what the likelihood of the update failing (if ever?) and building a new one. If it never fails then the flag can be ignored.

gfx_scene.h
maoliver-amd: yep ok done that

gfx.h
gboisse: Yeah, BVH updates in DXR are achieved through calling the build function indeed, but passing in the `D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PERFORM_UPDATE` flag.

So you'll most likely need to persist the flags passed by the user on build into the `RaytracingPrimitive` structure: https://github.com/gboisse/gfx/blob/master/gfx.h#L776-L788

I'm unsure what you mean with the update failing, but yes safe to assume it doesn't here I think 🙂
D3D12 being explicit, the only thing that can fail on an update would be outside the `BuildRaytracingAccelerationStructure()` and along the lines of failing to allocate enough memory for the BVH, the scratch, etc.
However, sufficient memory should already have been allocated on the initial build.

gboisse:
Oh and please remove the build artefacts from under the `build/` directory that are checked into this PR 🙂

maoliver-amd:
> Oh and please remove the build artefacts from under the `build/` directory that are checked into this PR 🙂

Not even sure how those snuck in :(

gfx.h
maoliver-amd: I did see the `RaytracingPrimitive` struct and was considering that but I was waiting for feedback. Ive added the flags to that in the most recent commit.

PR #23, opened 2022-04-26
Dump shader blobs (bytecodes and root signatures)

smeunier-amd:
[Radeon GPU Analyzer](https://gpuopen.com/learn/radeon-gpu-analyzer-2-2-direct3d12-compute/) can provide additional informations from DXIL using the command line.

These changes dump the shader bytecodes and the root signatures used by the application.

An example : 
```
rga.exe -s dx12 --cs-blob demo_UpdateShadows.csblob --rs-bin demo_UpdateShadows.rsbin --analysis analysis.txt
```

### DIFF
--- a/gfx.h
+++ b/gfx.h
@@ -3975,18 +3975,27 @@ public:
     }
 
 private:
-    // http://www.cse.yorku.ca/~oz/hash.html
-    static uint64_t Hash(char const *value)
-    {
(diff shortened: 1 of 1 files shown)


smeunier-amd:
We need a little more work for vertex and pixel shaders (some PSO descriptions).

PR #28, opened 2022-05-06
Ensure each animation only affect its animated nodes.

smeunier-amd:
I checked all animated scenes we use.


smeunier-amd:
I rebased on master.

PR #29, opened 2022-05-06
Ensure each animation only affect its animated nodes

smeunier-amd:
I checked all animated scenes we use.

### DIFF
--- a/gfx_scene.h
+++ b/gfx_scene.h
@@ -1663,14 +1663,14 @@ private:
                 animation_channel.type_ = type;
             }
         }
-        std::function<bool(int32_t, glm::dmat4 const &, bool)> VisitNode;
-        VisitNode = [&](int32_t node_index, glm::dmat4 const &parent_transform, bool is_parent_animated)
+        std::function<std::pair<bool, size_t>(int32_t, glm::dmat4 const &, bool, size_t)> VisitNode;
+        VisitNode = [&](int32_t node_index, glm::dmat4 const &parent_transform, bool is_parent_animated, size_t parent_animation) -> std::pair<bool, size_t>
         {
             glm::dvec3 T(0.0), S(1.0);
             glm::dquat R(1.0, 0.0, 0.0, 0.0);
             glm::dmat4 local_transform(1.0);    // default to identity
             if(node_index < 0 || node_index >= (int32_t)gltf_model.nodes.size())
-                return false;   // out of bounds
+                return {false, std::numeric_limits<size_t>::max()};   // out of bounds
             tinygltf::Node const &gltf_node = gltf_model.nodes[node_index];
             if(!gltf_node.matrix.empty())
             {
@@ -1744,12 +1744,23 @@ private:
                     TransformGltfLight(*light, transform);
                 }
             }
-            bool is_any_children_animated = false;
             std::map<int32_t, std::pair<size_t, uint64_t>>::const_iterator const it = animated_nodes.find(node_index);
             bool is_node_animated = (is_parent_animated || it != animated_nodes.end());
+            size_t node_animation = (is_parent_animated ? parent_animation : (it != animated_nodes.end() ? (*it).second.first : std::numeric_limits<size_t>::max()));
+            bool is_any_children_animated = false;
+            size_t children_animation = std::numeric_limits<size_t>::max();
             for(size_t i = 0; i < gltf_node.children.size(); ++i)
-                is_any_children_animated = (VisitNode(gltf_node.children[i], transform, is_node_animated) || is_any_children_animated);
-            is_node_animated = (is_node_animated || is_any_children_animated);
+            {
+                auto const [is_child_animated, child_animation] = VisitNode(gltf_node.children[i], transform, is_node_animated, node_animation);
+                GFX_ASSERT(!is_any_children_animated || !is_child_animated || children_animation == child_animation); // we assume non overlapping animations
+                is_any_children_animated = is_any_children_animated || is_child_animated;
+                children_animation = is_child_animated ? child_animation : children_animation;
+            }
+            if (!is_node_animated && is_any_children_animated)
+            {
+                is_node_animated = true;
+                node_animation = children_animation;
+            }
             if(is_node_animated)
             {
                 GltfNode *node = nullptr;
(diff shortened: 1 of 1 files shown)


gfx_scene.h
```diff
@@ -1602,7 +1602,7 @@ class GfxSceneInternal
             }
         }
         std::set<uint64_t> unparented_nodes;
-        std::map<int32_t, uint64_t> animated_nodes;
+        std::map<int32_t, std::pair<size_t, uint64_t>> animated_nodes;
```
smeunier-amd, line 1605: `animated_nodes` is shared between all animations but each animated node only belongs to one animation. So we store the animation index with the animated node handle.

gfx_scene.h
```diff
@@ -1744,35 +1744,47 @@ class GfxSceneInternal
...
+                auto const [is_child_animated, child_animation] = VisitNode(gltf_node.children[i], transform, is_node_animated, node_animation);
+                GFX_ASSERT(!is_any_children_animated || !is_child_animated || children_animation == child_animation); // we assume non overlapping animations
+                is_any_children_animated = is_any_children_animated || is_child_animated;
+                children_animation = is_child_animated ? child_animation : children_animation;
+            }
+            if (!is_node_animated && is_any_children_animated)
+            {
+                is_node_animated = true;
+                node_animation = children_animation;
+            }
```
smeunier-amd, line 1763: We propagate animations through the entire node hierarchy (from roots to leaves). We ensure each animated node (including parent nodes and children) is affected by one animation (no overlap between animation).

gfx_scene.h
```diff
@@ -1744,35 +1744,47 @@ class GfxSceneInternal
...
+                    gltf_node_refs_.insert(GetObjectIndex((*it).second.second), (uint32_t)animations.size());
                 }
                 else
                 {
+                    GFX_ASSERT(node_animation < animations.size());
                     uint64_t const animated_node_handle = gltf_node_handles_.allocate_handle();
                     gltf_node_refs_.insert(GetObjectIndex(animated_node_handle), (uint32_t)animations.size());
                     node = &gltf_nodes_.insert(GetObjectIndex(animated_node_handle));
-                    animated_nodes[node_index] = animated_node_handle;
+                    animated_nodes[node_index] = {node_animation, animated_node_handle};
```
smeunier-amd, line 1787: `node_animation` is an index, it can be the animation of the current node (if available) or the animation of one parent or child (if missing).

gfx_scene.h
```diff
@@ -1808,9 +1820,10 @@ class GfxSceneInternal
             GltfAnimation &animation_object = gltf_animations_[GetObjectIndex((*it).second)];
             for(size_t j = 0; j < gltf_scene.nodes.size(); ++j)
             {
-                std::map<int32_t, uint64_t>::const_iterator const it2 = animated_nodes.find(gltf_scene.nodes[j]);
-                if(it2 != animated_nodes.end())
-                    animation_object.nodes_.push_back((*it2).second);
+                std::map<int32_t, std::pair<size_t, uint64_t>>::const_iterator const it2 = animated_nodes.find(gltf_scene.nodes[j]);
+                GFX_ASSERT(it2 == animated_nodes.end() || (*it2).second.first < gltf_model.animations.size());
+                if(it2 != animated_nodes.end() && (*it2).second.first == i)
```
smeunier-amd, line 1825: We discard animated nodes belonging to another animation.

smeunier-amd: If we don't do this, each animated nodes will be updated by each animation. If we only play one animation at a time, it works but when a few or all animations are played, it can lead to weird results.

gfx_scene.h
```diff
@@ -66,7 +66,8 @@ class GfxRefBase { friend class GfxSceneInternal; protected: uint64_t handle; Gf
                    OBJECT_TYPE &operator *() const { return *operator ->(); }
                    operator bool() const { return !!operator ->(); }
                    operator uint32_t() const { return getIndex(); }
-                   operator uint64_t() const { return handle; } };
+                   operator uint64_t() const { return handle; }
+                   bool operator<(GfxRefBase const &other) const { return handle < other.handle; } };
```
smeunier-amd, line 70: For using `std::set`.

gfx_scene.h
```diff
@@ -476,7 +477,6 @@ class GfxSceneInternal
     };
 
     GfxArray<GltfNode> gltf_nodes_;
-    GfxArray<uint32_t> gltf_node_refs_;
```
smeunier-amd, line 479: I prefer using a set when visiting nodes for freeing handles one time.

gfx_scene.h
```diff
@@ -1663,14 +1665,18 @@ class GfxSceneInternal
                 animation_channel.type_ = type;
             }
         }
-        std::function<bool(int32_t, glm::dmat4 const &, bool)> VisitNode;
-        VisitNode = [&](int32_t node_index, glm::dmat4 const &parent_transform, bool is_parent_animated)
+        for(auto const& [_, animations] : node_animations)
+            if(animations.size() > 1)
+                GFX_PRINT_ERROR(kGfxResult_InternalError, "Some nodes are targeted by several animations...");
```
smeunier-amd, line 1670: Output a clear message when the scene contains overlapping animations.

gfx_scene.h
```diff
@@ -1744,37 +1750,40 @@ class GfxSceneInternal
...
+            {
+                std::map<int32_t, uint64_t>::const_iterator const it = animated_nodes.find(node_index);
                 GltfNode *node = nullptr;
-                std::vector<uint64_t> children;
                 GltfAnimatedNode *animated_node = nullptr;
-                for(size_t i = 0; i < gltf_node.children.size(); ++i)
-                {
-                    std::map<int32_t, uint64_t>::const_iterator const it2 = animated_nodes.find(gltf_node.children[i]);
-                    if(it2 != animated_nodes.end()) children.push_back((*it2).second);
-                }
```
smeunier-amd, line 1762: Moved a few lines after.

smeunier-amd:
I tested the scene with these changes.

maoliver-amd:
We currently have 2 buster_drones:
1) assets/busterDrone/busterDrone.gltf
2) assets/CapsaicinTestMedia/buster_drone/BusterDrone.gltf 

Its the second one thats broken

maoliver-amd:
> The ground is missing for some reasons.

Yeah it has a broken ground texture that has funky alpha values in it. This is fixed in the second buster scene (although apparently that has its own issues)

smeunier-amd:
It's a new case. Best option is to keep my changes and add a few more. :)

PR #30, opened 2022-05-09
Increase MaxNameLength

AlexanderVeselov:
32 symbols might not be enough in some cases


gboisse:
What about setting up for a middle ground of `64`?
Just to avoid making the structs too large 🙂

gboisse:
Closing due to inactivity.

PR #26, opened 2022-05-05
Fix failed assert introduced in "Link properly animated nodes and animations"

maoliver-amd:
Title pretty much says it all

PR #32, opened 2022-05-18
Fix light animation

maoliver-amd:
Firstly I changed the delta light transform to point toward the light instead of away (and fixed a bug)

Then ive updated the animation code as when there are multiple animations the current behavior assigns every detected animated node to every animation. This breaks animations as multiple incorrect animations are applied to each node wrongly attached to it. After spending alot of time digging around trying to understand what was going on I realised my fix is very similar to the one Sylvain did earlier that was reverted. So let me know if this version has any similar issues


gboisse:
The fix from Sylvain initially had the issues described over here: https://github.com/gboisse/gfx/pull/26#issuecomment-1118329560

But he's fixed the issues in https://github.com/gboisse/gfx/pull/29, which is now merged.

PR #42, opened 2022-08-08
Enable HLSL 2021 feature support.

maoliver-amd:
This enables templates and operator overloading support in hlsl.

It does requires converting uses of ```?:``` to the new ```select``` function but ive done that in all relevant locations.

### DIFF
--- a/gfx.h
+++ b/gfx.h
@@ -6222,7 +6222,7 @@ private:
         uint32_t const key = ((texture_type << 2) | channels);  // lookup key
         std::map<uint32_t, MipKernels>::const_iterator const it = mip_kernels_.find(key);
         if(it != mip_kernels_.end()) return (*it).second;   // already compiled
-        char const *texture_type_str = nullptr, *channel_type_str = nullptr, *did_type_str = nullptr;
+        char const *texture_type_str = nullptr, *channel_type_str = nullptr, *did_type_str = nullptr, *select_string = nullptr;
         switch(texture_type)
         {
         case GfxTexture::kType_2D:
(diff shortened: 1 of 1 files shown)


gfx.h
```diff
@@ -6303,13 +6303,13 @@ class GfxInternal
...
+        mip_program_source += "(select(";
         mip_program_source += val_string;
-        mip_program_source += " < 0.04045f ? ";
+        mip_program_source += " < 0.04045f, ";
         mip_program_source += val_string;
-        mip_program_source += " / 12.92f : pow((";
+        mip_program_source += " / 12.92f, pow((";
         mip_program_source += val_string;
-        mip_program_source += " + 0.055f) / 1.055f, 2.4f)";
+        mip_program_source += " + 0.055f) / 1.055f, 2.4f))";
```
gboisse, line 6312: Mmh, this code now using the new `select()` intrinsic will now fail if HLSL 2021 isn't supported, right?

gboisse (changes requested):

gfx.h
maoliver-amd: Yep its only 2021 or newer. Line 7161 below enables HLSL2021 as standard and gfx bundles a compatible version of dxc (which is the only thing that touches the HLSL2021 changes as the DXIR passed to the driver hasnt changed) so it would never fail aslong as everything was still within gfx (or something using gfx)

maoliver-amd: So the only breaking change would be for any downstream user of gfx that has there own shaders that they are using gfx to compile but are not compatible with the new ```select``` requirement. However, the shader compilation output log will clearly state the need for using ```select``` and where it is required, it just unfortunately will require any user that falls into that category to make the required changes.

maoliver-amd: I added the pre-HLSL2021 select function to the mip shader code just in case.

gboisse: Ok perfect, thanks!

gboisse (approved):

PR #44, opened 2022-08-22
Texture compression support in gltf

maoliver-amd:
This adds basisu texture compressions and transcoding support for gltf.

In order to support the required basisu extension I ended up having to change the gltf library from tinygltf to cgltf (which also supports other extensions that can be easily enabled as well). This change actually notably simplified the gltf loading code while I tried to keep it similar to what was already there.

I also implemented a non-standard feature in that when a metallicity-roughness texture is found in the gltf file it will look for optional standalone (i.e. separate) texture files for each component. There is discussion about adding this sort of capability as an official extension, as combining metal/rough when using texture compression is not ideal, but it will be a little while before that eventuates so this is an interim workaround.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,16 @@
 [submodule "third_party/tinyobjloader-2.0-rc1"]
 	path = third_party/tinyobjloader-2.0-rc1
 	url = https://github.com/tinyobjloader/tinyobjloader.git
+[submodule "third_party/cgltf"]
+	path = third_party/cgltf
+	url = https://github.com/jkuhlmann/cgltf.git
+[submodule "third_party/KTX-Software"]
+	path = third_party/KTX-Software
+	url = https://github.com/KhronosGroup/KTX-Software.git
+[submodule "third_party/Vulkan-Headers"]
+	path = third_party/Vulkan-Headers
+	url = https://github.com/KhronosGroup/Vulkan-Headers.git
+[submodule "third_party/stb"]
+	path = third_party/stb
+	url = https://github.com/nothings/stb.git
+	branch = master
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,6 +10,3 @@
 [submodule "third_party/tinyobjloader-2.0-rc1"]
 	path = third_party/tinyobjloader-2.0-rc1
 	url = https://github.com/tinyobjloader/tinyobjloader.git
-[submodule "third_party/tinygltf-2.5.0"]
-	path = third_party/tinygltf-2.5.0
-	url = https://github.com/syoyo/tinygltf.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,15 +1,15 @@
-[submodule "third_party/imgui-1.79"]
-	path = third_party/imgui-1.79
-	url = https://github.com/ocornut/imgui.git
-[submodule "third_party/D3D12MemoryAllocator-1.0.0"]
-	path = third_party/D3D12MemoryAllocator-1.0.0
-	url = https://github.com/GPUOpen-LibrariesAndSDKs/D3D12MemoryAllocator.git
-[submodule "third_party/glm-0.9.9.8"]
-	path = third_party/glm-0.9.9.8
-	url = https://github.com/g-truc/glm.git
-[submodule "third_party/tinyobjloader-2.0-rc1"]
-	path = third_party/tinyobjloader-2.0-rc1
+[submodule "third_party/tinyobjloader"]
+	path = third_party/tinyobjloader
 	url = https://github.com/tinyobjloader/tinyobjloader.git
+[submodule "third_party/imgui"]
+	path = third_party/imgui
+	url = https://github.com/ocornut/imgui.git
+[submodule "third_party/D3D12MemoryAllocator"]
+	path = third_party/D3D12MemoryAllocator
+	url = https://github.com/GPUOpen-LibrariesAndSDKs/D3D12MemoryAllocator.git
+[submodule "third_party/glm"]
+	path = third_party/glm
+	url = https://github.com/g-truc/glm.git
 [submodule "third_party/cgltf"]
 	path = third_party/cgltf
 	url = https://github.com/jkuhlmann/cgltf.git
--- a/gfx.h
+++ b/gfx.h
@@ -2758,24 +2758,17 @@ public:
         D3D12_PLACED_SUBRESOURCE_FOOTPRINT subresource_footprints[D3D12_REQ_MIP_LEVELS] = {};
         device_->GetCopyableFootprints(&resource_desc, 0, dst.mip_levels, 0, subresource_footprints, num_rows, row_sizes, nullptr);
         uint64_t buffer_offset = 0;
-        uint32_t pixels_per_block = 1;
         uint32_t const bytes_per_pixel = GetBytesPerPixel(dst.format);
         if(bytes_per_pixel == 0)
             return GFX_SET_ERROR(kGfxResult_InvalidOperation, "Cannot copy to texture object of unsupported format");
-        if(dst.format >= DXGI_FORMAT_BC1_TYPELESS && dst.format <= DXGI_FORMAT_BC5_SNORM)
-        {
-            pixels_per_block = 4 * 4;   // BC formats have 4*4 pixels per block
-            pixels_per_block /= 4;  // we need to divide by 4 because GetCopyableFootprints introduces a *2 stride divides the rows /4
-        }
         for(uint32_t mip_level = 0; mip_level < dst.mip_levels; ++mip_level)
         {
             Buffer *texture_upload_buffer = nullptr;
             if(buffer_offset >= src.size)
                 break;  // further mips aren't available
-            GFX_ASSERT(num_rows[mip_level] == GFX_MAX(dst.height >> mip_level, 1u));
-            uint32_t const buffer_row_pitch = (GFX_MAX(dst.width >> mip_level, 1u) * bytes_per_pixel) / pixels_per_block;
-            uint32_t const texture_row_pitch = subresource_footprints[mip_level].Footprint.RowPitch;
-            uint64_t const buffer_size = GFX_MAX(dst.height >> mip_level, 1u) * buffer_row_pitch;
+            uint64_t const buffer_row_pitch = row_sizes[mip_level];
+            uint64_t const texture_row_pitch = subresource_footprints[mip_level].Footprint.RowPitch;
+            uint64_t const buffer_size = (uint64_t)num_rows[mip_level] * buffer_row_pitch;
             if(buffer_offset + buffer_size > src.size)
                 return GFX_SET_ERROR(kGfxResult_InvalidOperation, "Cannot copy to mip level %u from buffer object with insufficient storage", mip_level);
             if(buffer_row_pitch != texture_row_pitch || // we must respect the 256-byte pitch alignment
@@ -2799,7 +2792,7 @@ public:
                 transitionResource(*texture_upload_buffer, D3D12_RESOURCE_STATE_COPY_DEST);
                 submitPipelineBarriers();   // transition our resources if needed
                 for(uint32_t i = 0; i < num_rows[mip_level]; ++i)
-                    command_list_->CopyBufferRegion(texture_upload_buffer->resource_, i * texture_row_pitch, gfx_buffer.resource_, i * buffer_row_pitch, buffer_row_pitch);
+                    command_list_->CopyBufferRegion(texture_upload_buffer->resource_, i * texture_row_pitch, gfx_buffer.resource_, i * buffer_row_pitch + buffer_offset, buffer_row_pitch);
             }
             Buffer &gfx_buffer = buffers_[src]; SetObjectName(gfx_buffer, src.name);
             if(texture_upload_buffer != nullptr) transitionResource(*texture_upload_buffer, D3D12_RESOURCE_STATE_COPY_SOURCE);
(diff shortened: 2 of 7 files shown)


.gitmodules
```diff
@@ -10,6 +10,16 @@
...
+[submodule "third_party/KTX-Software"]
+	path = third_party/KTX-Software
+	url = https://github.com/KhronosGroup/KTX-Software.git
+[submodule "third_party/Vulkan-Headers"]
+	path = third_party/Vulkan-Headers
+	url = https://github.com/KhronosGroup/Vulkan-Headers.git
+[submodule "third_party/stb"]
+	path = third_party/stb
+	url = https://github.com/nothings/stb.git
+	branch = master
```
maoliver-amd, line 25: Changed to using cgltf for gltf loading as it supports the required extensions. Since tinygltf is removed then we need a standalone version of stb (as previously it was in tinygltf). Ktx is needed to open and transcode ktx files which also requires the vulkan headers to identify the internal format of each texture (side effect of ktx being Khronos)

gfx.h
```diff
@@ -2758,24 +2758,17 @@ class GfxInternal
...
             Buffer *texture_upload_buffer = nullptr;
             if(buffer_offset >= src.size)
                 break;  // further mips aren't available
-            GFX_ASSERT(num_rows[mip_level] == GFX_MAX(dst.height >> mip_level, 1u));
-            uint32_t const buffer_row_pitch = (GFX_MAX(dst.width >> mip_level, 1u) * bytes_per_pixel) / pixels_per_block;
-            uint32_t const texture_row_pitch = subresource_footprints[mip_level].Footprint.RowPitch;
-            uint64_t const buffer_size = GFX_MAX(dst.height >> mip_level, 1u) * buffer_row_pitch;
+            uint64_t const buffer_row_pitch = row_sizes[mip_level];
+            uint64_t const texture_row_pitch = subresource_footprints[mip_level].Footprint.RowPitch;
+            uint64_t const buffer_size = (uint64_t)num_rows[mip_level] * buffer_row_pitch;
```
maoliver-amd, line 2771: Directly use ```row_sizes``` and ```subresource_footprints``` for determining the row size and this auto handles compressed textures as we are setting the required format in the resource description

gfx_scene.h
```diff
@@ -244,22 +245,22 @@ GfxMetadata const &gfxSceneGetImageMetadata(GfxScene scene, uint64_t image_handl
 //! Image helpers.
 //!
 
-inline DXGI_FORMAT gfxImageGetFormat(GfxImage const &image)
```
maoliver-amd, line 247: Made ```gfxImageGetFormat``` private as there is already a member variable of ```GfxImage``` that should be probed for the internal format (as with compressed formats the possible formats gets significantly larger than what this function could correctly handle.

gfx_scene.h
```diff
@@ -244,22 +245,22 @@ GfxMetadata const &gfxSceneGetImageMetadata(GfxScene scene, uint64_t image_handl
...
+        image.format == DXGI_FORMAT_BC2_UNORM || image.format == DXGI_FORMAT_BC2_UNORM_SRGB ||
+        image.format == DXGI_FORMAT_BC3_TYPELESS || image.format == DXGI_FORMAT_BC3_UNORM ||
+        image.format == DXGI_FORMAT_BC3_UNORM_SRGB || image.format == DXGI_FORMAT_BC4_TYPELESS ||
+        image.format == DXGI_FORMAT_BC4_UNORM || image.format == DXGI_FORMAT_BC4_SNORM ||
+        image.format == DXGI_FORMAT_BC5_TYPELESS || image.format == DXGI_FORMAT_BC5_UNORM ||
+        image.format == DXGI_FORMAT_BC5_SNORM || image.format == DXGI_FORMAT_BC6H_TYPELESS ||
+        image.format == DXGI_FORMAT_BC6H_UF16 || image.format == DXGI_FORMAT_BC6H_SF16 ||
+        image.format == DXGI_FORMAT_BC7_TYPELESS || image.format == DXGI_FORMAT_BC7_UNORM ||
+        image.format == DXGI_FORMAT_BC7_UNORM_SRGB)             {
+        return true;
```
maoliver-amd, line 261: Added an additional public function to check if a format is compressed or not (which is useful in various places)

gfx_scene.h
```diff
@@ -1110,473 +1166,267 @@ class GfxSceneInternal
...
                     {
+                        if (gfxSceneImport(scene, metallicity_map_file.c_str()) == kGfxResult_NoError)
+                            metallicity_map_ref = gfxSceneFindObjectByAssetFile<GfxImage>(scene, metallicity_map_file.c_str());
+                    }
+                    if (!metallicity_map_ref)
+                    {
+                        if (gfxSceneImport(scene, roughness_map_file.c_str()) == kGfxResult_NoError)
+                            roughness_map_ref = gfxSceneFindObjectByAssetFile<GfxImage>(scene, roughness_map_file.c_str());
+                    }
+                    if(!metallicity_map_ref && !roughness_map_ref)
```
maoliver-amd, line 1311: This is were we check for standalone already split versions of a image. So if the gltf file specifies a 'xxx.ext' image file it now first checks for a 'xxx.metallicity.ext' and/or 'xxx.roughness.ext' image and loads those directly instead (useful for compressed textures as we cant really separate the channels easily and having them already in separate images allows for each image to use a single channel BC4 compression which results in smaller files/memory and better quality). If no separate images are found then it does the manually splitting as before.

maoliver-amd: Since its possible for a material to only require a single texture for metallicity/roughness (as the other is just a constant) then we only try and split if both images are missing

gfx_scene.h
```diff
@@ -1110,473 +1166,267 @@ class GfxSceneInternal
...
+            cgltf_pbr_metallic_roughness const &gltf_material_pbr = gltf_material.pbr_metallic_roughness;
+            material.albedo = glm::vec4(gltf_material_pbr.base_color_factor[0], gltf_material_pbr.base_color_factor[1],
+                gltf_material_pbr.base_color_factor[2], gltf_material_pbr.base_color_factor[3]);
+            material.roughness = gltf_material_pbr.roughness_factor;
+            material.metallicity = gltf_material_pbr.metallic_factor;
+            material.emissivity = glm::vec3(gltf_material.emissive_factor[0], gltf_material.emissive_factor[1],
+                gltf_material.emissive_factor[2]);
+            cgltf_texture const *albedo_map_text = gltf_material_pbr.base_color_texture.texture;
+            it = (albedo_map_text != nullptr ? images.find(albedo_map_text->basisu_image != nullptr ?
+                albedo_map_text->basisu_image : albedo_map_text->image) : images.end());
```
maoliver-amd, line 1271: Since compressed textures are an option extension it is possible for a gltf file to have both a compressed image (```basisu_image```) and a normal image (```image```). So since we support compressed we use that first if it is specified otherwise load the regular image.

gfx_scene.h
```diff
@@ -1886,39 +1716,170 @@ class GfxSceneInternal
...
     GfxResult importImage(GfxScene const &scene, char const *asset_file)
     {
         GFX_ASSERT(asset_file != nullptr);
         int32_t image_width, image_height, channel_count;
         if(gfxSceneFindObjectByAssetFile<GfxImage>(scene, asset_file))
             return kGfxResult_NoError;  // image was already imported
-        stbi_uc *image_data = stbi_load(asset_file, &image_width, &image_height, &channel_count, 0);
+        stbi_uc *image_data = nullptr;
+        uint32_t bytes_per_channel = 2;
+        if(stbi_is_16_bit(asset_file))
```
maoliver-amd, line 1833: Also added support for loading 16bit images. Previously the tinygltf loader had 16b image support but this makes the functionality available to all.

gfx_scene.h
```diff
@@ -1886,39 +1716,170 @@ class GfxSceneInternal
...
+                for (int32_t x = 0; x < image_width; ++x)
+                    for (int32_t k = 0; k < (int32_t)resolved_channel_count; ++k) {
+                        int32_t const dst_index = (int32_t)resolved_channel_count * (x + y * image_width) + k;
+                        int32_t const src_index = channel_count * (x + y * image_width) + k;
+                        uint16_t const source = (k < channel_count ? image_data[src_index] : (uint16_t)65535);
+                        if (k == 3) alpha_check &= source;
+                        data[dst_index] = source;
+                    }
+            image_ref->flags = (alpha_check == 65535 ? 0 : kGfxImageFlag_AlphaChannel);
+        }
```
maoliver-amd, line 1882: Loading image data now supports both 8b and 16b. The main difference though is that this loads images in the order that the data is contained in the source file, whereas previously it was loaded flipped. This new version is what is required for loading images as found in gltf files but might cause issues with obj files being flipped????

.gitmodules
gboisse: Pretty happy about that removal of `tinygltf` tbh, it's only "tiny" in the name really...
Did you notice any performance speed-up (or otherwise) with this change?

gfx.h
```diff
@@ -2799,7 +2792,7 @@ class GfxInternal
                 transitionResource(*texture_upload_buffer, D3D12_RESOURCE_STATE_COPY_DEST);
                 submitPipelineBarriers();   // transition our resources if needed
                 for(uint32_t i = 0; i < num_rows[mip_level]; ++i)
-                    command_list_->CopyBufferRegion(texture_upload_buffer->resource_, i * texture_row_pitch, gfx_buffer.resource_, i * buffer_row_pitch, buffer_row_pitch);
+                    command_list_->CopyBufferRegion(texture_upload_buffer->resource_, i * texture_row_pitch, gfx_buffer.resource_, i * buffer_row_pitch + buffer_offset, buffer_row_pitch);
```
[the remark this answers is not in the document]
gboisse: Oh snap, good spotting.

gfx_scene.h
gboisse: Agreed.

gfx_scene.h
gboisse: Sounds good, really wish MS would supply some ready-made helpers/macros for these...

gfx_scene.h
```diff
@@ -1110,337 +1166,155 @@ class GfxSceneInternal
...
+                continue;
+            cgltf_image const *gltf_image = gltf_texture.has_basisu ? gltf_texture.basisu_image : gltf_texture.image;
+            GfxRef<GfxImage> image_ref;
+            if(gltf_image->uri == nullptr) continue; // no image to be loaded
+            char const *filename = GFX_MAX(strrchr(asset_file, '/'), strrchr(asset_file, '//'));
+            std::string image_folder = (filename != nullptr ? std::string(asset_file, 0, filename - asset_file) : std::string(""));
+            if(!image_folder.empty() && image_folder.back() != '/' && image_folder.back() != '\\') image_folder += '/';
+            std::string image_file = image_folder + gltf_image->uri;
+            if(gfxSceneImport(scene, image_file.c_str()) != kGfxResult_NoError)
+                continue; // unable to load image file
```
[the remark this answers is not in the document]
gboisse: Great.

gfx_scene.h
gboisse: I believe the issue is that obj's UV vertical origin is swapped compared to glTF's configuration.
So like for one format y=0 means top, for the other y=0 means bottom...
That's also why the image loading was "repeated" inside of the previous `glTF` loading code iirc.
One option could be adding a "flip_image" option to the image loading function? or try and invert the obj's UVs somehow?

gboisse: Just the vertical component of the UVs obviously.

gboisse: I think just having a `uv.y = 1.0f - uv.y` in the obj loader might suffice to fix the issue, bit annoying but it does seem like the less intrusive solution.

gfx_scene.h
gboisse: Sounds good.

gboisse:
It's a bit hard for me to review this just by looking at the code 😅
So I propose that we merge this once the flipped obj textures issue is resolved, then I'll go and test it on content.
Really great that we're getting compressed textures support though, we'll be able to test way bigger scenes now 🙂

.gitmodules
maoliver-amd: Yes, cgltf is slightly faster than tinygltf in what ive seen.

gfx_scene.h
maoliver-amd: OK, ill flip the uvs in the obj loader as I think thats probably the best solution (as it allows compressed textures to be used there as well)

gfx_scene.h
maoliver-amd: Well there is now the ```gfxImageIsFormatCompressed``` function that is added in place of the removed ```gfxImageGetFormat``` which still classifies as an image helper???

gboisse: Oh right I see, then it's certainly good!
Wasn't visible from the small github previous 😅

maoliver-amd: Yeah it shows up down below somewhere :smile:

gfx_scene.h
gboisse: Ok great, let me know when this is done.

gfx_scene.h
```diff
@@ -1057,7 +1113,7 @@ class GfxSceneInternal
                         if(std::get<2>(key) >= 0)
                         {
                             vertex.uv.x = obj_reader.GetAttrib().texcoords[2 * std::get<2>(key) + 0];
-                            vertex.uv.y = obj_reader.GetAttrib().texcoords[2 * std::get<2>(key) + 1];
+                            vertex.uv.y = 1.0f - obj_reader.GetAttrib().texcoords[2 * std::get<2>(key) + 1];
```
maoliver-amd, line 1116: flipped the uvs in .obj files to allow them to be used with the default image loading

gfx_scene.h
maoliver-amd: Added the fix (seems pretty simple)

PR #45, opened 2022-08-25
Update submodules

maoliver-amd:
Updates submodules to latest version with any required changes.

D3D12MemoryAllocator: v2.0.1
imgui: v1.62
tinyobjloader: v2.0.0rc9

### DIFF
--- a/gfx.h
+++ b/gfx.h
@@ -454,6 +454,7 @@ class GfxInternal
     uint32_t max_frames_in_flight_ = 0;
 
     ID3D12Device *device_ = nullptr;
+    IDXGIAdapter1 *adapter_ = nullptr;
     ID3D12Device5 *dxr_device_ = nullptr;
     ID3D12CommandQueue *command_queue_ = nullptr;
     ID3D12GraphicsCommandList *command_list_ = nullptr;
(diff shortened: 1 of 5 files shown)


gboisse:
Mmh the version is the names is a bit on purpose so we can immediately see what version is being used for each dependency.
It does make the updating slightly more painful I suppose, but I also think it makes it much clearer/more explicit what's the state of the dependencies?

gfx.h
```diff
@@ -454,6 +454,7 @@ class GfxInternal
     uint32_t max_frames_in_flight_ = 0;
 
     ID3D12Device *device_ = nullptr;
+    IDXGIAdapter1 *adapter_ = nullptr;
```
maoliver-amd, line 457: ```D3D12MA::CreateAllocator``` now requires that an adapter also be passed along with the device so I added the adapter as an internal variable

gfx.h
```diff
@@ -1138,38 +1139,26 @@ class GfxInternal
...
-
-            struct DXGIAdapterReleaser
-            {
-                IDXGIAdapter1 *adapter;
-                GFX_NON_COPYABLE(DXGIAdapterReleaser);
-                DXGIAdapterReleaser(IDXGIAdapter1 *adapter) : adapter(adapter) {}
-                ~DXGIAdapterReleaser() { if(adapter) adapter->Release(); }
-            };
-            DXGIAdapterReleaser const adapter_releaser(adapter1);
-
```
maoliver-amd, line 1154: No longer any need for adapter descriptions as theres now the adapter stored directly internally so this was all removed

gfx_imgui.h
```diff
@@ -159,28 +160,6 @@ class GfxImGuiInternal
...
-        io.KeyMap[ImGuiKey_Space]      = VK_SPACE;
-        io.KeyMap[ImGuiKey_Enter]      = VK_RETURN;
-        io.KeyMap[ImGuiKey_Escape]     = VK_ESCAPE;
-        io.KeyMap[ImGuiKey_A]          = 0x41;  // https://docs.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes
-        io.KeyMap[ImGuiKey_C]          = 0x43;
-        io.KeyMap[ImGuiKey_V]          = 0x56;
-        io.KeyMap[ImGuiKey_X]          = 0x58;
-        io.KeyMap[ImGuiKey_Y]          = 0x59;
-        io.KeyMap[ImGuiKey_Z]          = 0x5A;
-
```
maoliver-amd, line 183: New imgui versions have a new io API that replaces the KepMap and KeysDown API.

gfx_imgui.h
```diff
@@ -319,12 +298,13 @@ class GfxImGuiInternal
 
         io.MouseWheel  = mouse_wheel;
         io.MouseWheelH = mouse_wheelh;
-        if(g_hWnd != 0)
+        ImGui_ImplWin32_Data* bd = ImGui_ImplWin32_GetBackendData();
+        if(bd != nullptr && bd->hWnd != 0)
```
maoliver-amd, line 302: Theres no global ```g_hWnd``` anymore in the win32 backend as everything is now encapsulated inside ```ImGui_ImplWin32_Data```

gfx_window.h
```diff
@@ -135,7 +134,7 @@ class GfxWindowInternal
     {
         if(window_)
             DestroyWindow(window_);
-        if(is_imgui_initialized_)
+        if(ImGui_ImplWin32_GetBackendData() != nullptr)
```
maoliver-amd, line 137: There was an issue with ```is_imgui_initialized_ getting``` inside gfx_window getting out of sync with gfx_imgui. ```ImGui_ImplWin32_GetBackendData()``` prevents this as it is always null if imgui has not been initialised. Using this directly prevents a crash that occurs occasionally when gfxImgui*** functions are called resulting in a imgui shutdown that gfx_windows isnt aware of.

maoliver-amd:
> Mmh the version is the names is a bit on purpose so we can immediately see what version is being used for each dependency. It does make the updating slightly more painful I suppose, but I also think it makes it much clearer/more explicit what's the state of the dependencies?

Yeah I figured as much. Just running ```git submodule``` from the gfx folder though also gives the complete list of submodule versions. So Im not sure which results in less work, running ```git submodule``` to check the current versions or renaming the folders everytime the submodule is updated.

The names are just the default here as in the texture_compression PR the names were just using the default (as I didnt change them from what ```git submodule add``` makes so I just did the same thing here (actually deleted the older folders and made new ones using the default as git went a little silly with whatever I did when I renamed the existing folder)

gfx.h
gboisse: Sounds good.

gfx_window.h
gboisse: Ah, that does sounds like a good change indeed, maybe we can get rid of that `is_imgui_initialized_` altogether then?

gboisse:
> Yeah I figured as much. Just running git submodule from the gfx folder though also gives the complete list of submodule versions. So Im not sure which results in less work, running git submodule to check the current versions or renaming the folders everytime the submodule is updated.

Yea, you're right, didn't know about this.
The version numbers don't even match although I did use the commit tagged for each release 😄

gboisse (approved):

gfx_window.h
maoliver-amd: I did ;) (line 81)

PR #46, opened 2022-08-30
Fix mouse wheel behaviour.

maoliver-amd:
Seems to do the job


gboisse:
lol, code removal is the best kind of fix 🙂

PR #48, opened 2022-09-15
Add additional material properties.

maoliver-amd:
Adds support for:

- Specifying IOR directly for fresnel calculations
- Specular F0 term
- Transmission weight
- Sheen
- Clearcoat

All added material variables correspond to the GLTF specifications and as such not everything can be mapped from an imported obj (which only has ior and specular added). I didnt add iridescence support due to its complexity and skipped the ability to define a second normal map just for the clearcoat (not something blender supports anyway)


gboisse:
Nice additions, thanks! 🙂

PR #50, opened 2022-09-19
Fix for missing names in gltf loading.

### LINKED ISSUE #49: New glTF importer crashes when objects have no valid name
So, it seems that `cgltf` returns `nullptr` names on objects that have no valid name, which crashes `std::string` as they do not accept `nullptr` assignments (which is a really annoying thing if you ask me): https://github.com/gboisse/gfx/commit/c31f5ceebf2b54a83c3c982cde0c6b6c5865fc81 (could cause the crash by trying to load the Sponza scene)

This ticket is mostly to track whether you think there's other places where this check should be added?

maoliver-amd:
Fixes #49 

This should cover everything now, I changed the material name to match everywhere else where it creates a default name.

### DIFF
--- a/gfx_scene.h
+++ b/gfx_scene.h
@@ -1294,7 +1294,7 @@ private:
             GfxRef<GfxMaterial> material_ref = gfxSceneCreateMaterial(scene);
             GfxMetadata &material_metadata = material_metadata_[material_ref];
             material_metadata.asset_file = asset_file;  // set up metadata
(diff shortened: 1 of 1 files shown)


gboisse:
Thanks, perfect!
Tbh, I'm quite annoyed that `std::string` crashes on `nullptr`, who in their right mind would think that's a good design decision? 😛

PR #54, opened 2022-10-04
Add dxcompiler and WinPix libraries as cmake targets.

maoliver-amd:
This allows for cmake to auto propagate the dependencies to any project using gfx which will then auto install the required dll's without requiring custom copy commands.


gboisse:
Nice, didn't know this was possible.

PR #55, opened 2022-10-06
Fix compilation with msvc c++20.

maoliver-amd:
GFX doesnt currently compile with C++20 with msvc compiler.
This PR fixes that with 2 changes:
 - Needed ```this->``` added to handle to get it correctly picked up as the class member variable
 - Added ```std::is_void_v<TYPE>``` to cause delayed instantiation of default template as otherwise the static_assert will always trigger (The choice of ```is_void_v``` is somewhat random as all it needs is some way to delay template evaluation and void type should never occur anyway.

### DIFF
--- a/gfx_scene.h
+++ b/gfx_scene.h
@@ -74,8 +74,8 @@ template<typename TYPE>
 class GfxRef : public GfxRefBase<TYPE, TYPE> { friend class GfxConstRef<TYPE>; };
 
 template<typename TYPE>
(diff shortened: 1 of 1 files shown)


gboisse:
Great, thanks.
I understand the removal of the `static_assert(false, ...)` but really surprised about the `this->` being required?
Do you know why it's needed in this particular situation?
Certainly hope we're not going to have to start adding that to all member variables with future versions of C++ 😛

maoliver-amd:
Its because c++20 adds some new features, one of which is named handle. So it just requires the ```this->``` to differentiate it to the compiler. So luckily not something that has to be added to everything 😄

PR #59, opened 2022-11-15
Add gltf support for sparse accessors and non indexed meshes.

maoliver-amd:
Adds support for sparse buffer accessors
Fixes loading meshes that dont define an index buffer (it actually just defines an default index buffer now - this was actually easier than not creating an index buffer and makes it easier to add support for triangle strips/fans later)
Fixes loading files that dont define a default scene.


gboisse:
Thanks for the fixes.

PR #60, opened 2022-11-17
Fix nested animations.

maoliver-amd:
Fixes issues where node transforms are not properly propagated through the node hierarchy.


gfx_scene.h
```diff
@@ -1773,14 +1765,23 @@ class GfxSceneInternal
...
+                propogate_parent_animations.insert(
+                    node_animations[gltf_node].begin(), node_animations[gltf_node].end());
+                children_animations.insert(
+                    node_animations[gltf_node].begin(), node_animations[gltf_node].end());
+            }
             for(size_t i = 0; i < gltf_node->children_count; ++i)
             {
-                auto child_animations = VisitNode(gltf_node->children[i], transform, node_animations[gltf_node]);
+                auto child_animations =
+                    VisitNode(gltf_node->children[i], transform, propogate_parent_animations);
```
maoliver-amd, line 1779: Need to propogate the current node transform as well as the parent tansform to the children otherwise parent animations are lost and not all child nodes are created

gfx_scene.h
```diff
@@ -1804,13 +1805,18 @@ class GfxSceneInternal
                 for(size_t i = 0; i < gltf_node->children_count; ++i)
                 {
                     std::map<cgltf_node const *, uint64_t>::const_iterator const it2 = animated_nodes.find(gltf_node->children[i]);
-                    if(it2 != animated_nodes.end()) children.push_back((*it2).second);
+                    GFX_ASSERT(it2 != animated_nodes.end());
+                    children.push_back((*it2).second);
```
maoliver-amd, line 1809: All children of an animated parent should create a node for animation

gfx_scene.h
```diff
@@ -1804,13 +1805,18 @@ class GfxSceneInternal
...
-                    animated_node->scale_ = S;
-                    animated_node->rotate_ = R;
-                    animated_node->translate_ = T;
+                    animated_node->translate_ = local_transform[3];
+                    for (int i = 0; i < 3; i++)
+                        animated_node->scale_[i] = glm::length(glm::dvec3(local_transform[i]));
+                    const glm::dmat3 rotMtx(glm::dvec3(local_transform[0]) / animated_node->scale_[0],
+                        glm::dvec3(local_transform[1]) / animated_node->scale_[1],
+                        glm::dvec3(local_transform[2]) / animated_node->scale_[2]);
+                    animated_node->rotate_ = glm::quat_cast(rotMtx);
```
maoliver-amd, line 1819: Nodes that had a fixed transform as well as an animated transform where nor correctly being set. This adds the fixed transform parts.

gboisse: Great fix, thanks.

PR #71, opened 2023-06-14
ImGui: Support specifying multiple fonts with attached font configs

maoliver-amd:
ImGui by default only loads glyphs for standard Latin characters (i.e. first 255). In order to load glyphs outside this range you need to pass in the requested range using a font config struct.
This PR allows specifying a font config struct with each font (now also supports passing multiple fonts)

example usage:
```
char const  *fonts[] = {"C:\\Windows\\Fonts\\seguisym.ttf"};
ImFontConfig fontConfigs[1];
fontConfigs[0].MergeMode           = true;
static const ImWchar glyphRanges[] = {
    0x23E0,
    0x23FF, // Media player icons
    0x1F500,
    0x1F505, // Restart icon
    0,
};
fontConfigs[0].GlyphRanges = &glyphRanges[0];
fontConfigs[0].SizePixels  = 30.0f;
fontConfigs[0].GlyphOffset.y += 5.0f; // Need to offset glyphs downward to properly center them
auto err = gfxImGuiInitialize(contextGFX, fonts, 1, fontConfigs);
```

### DIFF
--- a/gfx_imgui.h
+++ b/gfx_imgui.h
@@ -25,13 +25,15 @@ SOFTWARE.
 #define GFX_INCLUDE_GFX_IMGUI_H
 
 #include "gfx.h"
+#define IMGUI_USE_WCHAR32
 #include "imgui.h"
 
 //!
 //! ImGui initialization/termination.
 //!
 
-GfxResult gfxImGuiInitialize(GfxContext gfx, char const *font_filename = nullptr, ImGuiConfigFlags flags = 0);
+GfxResult gfxImGuiInitialize(GfxContext gfx, char const **font_filename = nullptr, uint32_t font_count = 0,
+    ImFontConfig const *font_configs = nullptr, ImGuiConfigFlags flags = 0);
 GfxResult gfxImGuiTerminate();
 GfxResult gfxImGuiRender();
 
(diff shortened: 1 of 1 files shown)


gfx_imgui.h
```diff
@@ -25,13 +25,15 @@ SOFTWARE.
 #define GFX_INCLUDE_GFX_IMGUI_H
 
 #include "gfx.h"
+#define IMGUI_USE_WCHAR32
```
maoliver-amd, line 28: Required to support unicode values above 0x10000

gfx_imgui.h
```diff
@@ -90,8 +92,15 @@ class GfxImGuiInternal
 
         uint8_t *font_data;
         int32_t font_width, font_height;
-        if(font_filename != nullptr)
-            io.Fonts->AddFontFromFileTTF(font_filename, 16.0f);
+        for (uint32_t i = 0; i < font_count; ++i)
+        {
+            if (i == 0 && font_configs != nullptr && font_configs[0].MergeMode == true)
+				io.Fonts->AddFontDefault();
```
maoliver-amd, line 98: If the first user font has configuration set to merge then we load in the default font and then merge the new fonts glyphs into the same glyph atlas. This allows adding additional glyphs to the default font as that only supports basic Latin characters.
If not set to merge then the new font will become the new default

PR #74, opened 2023-07-13
Sbt descriptor heap consistency

skkkksdkfak:


gfx.h
```diff
@@ -5900,6 +5862,74 @@ class GfxInternal
...
+                            }
+                        }
+                    }
+                    command_list_->CopyBufferRegion(sbt_buffer.resource_, dst_offset, upload_buffer.resource_,
+                        src_offset, upload_buffer_offset - src_offset);
+                    sbt_record.commited_id_ = sbt_record.id_;
+                }
+                destroyBuffer(upload_gfx_buffer);
+            }
+        }
```
gboisse, line 5932: This seems to be a copy/paste of the block above.

It probably would be best to factor this, what do you think?

skkkksdkfak: Indeed, the second block looks pretty much similar to previous one, but we can't merge them into just one loop, as the first loop intended to allocate descriptors, and if during the allocation we ran out of free descriptors, we allocate a new descriptor heap and populate it in the second loop, actually it inherits logic from regular descriptor binding population, just for SBT.

PR #79, opened 2023-09-06
Make scene and UI dependencies optional

maoliver-amd:
Adds 2 cmake configuration options `GFX_ENABLE_GUI` and `GFX_ENABLE_SCENE` which default to disabled. The first option enables imgui support while the second adds all the dependencies needed for gfx_scene (glm, stb, cgltf, tinyobjloader, tinyexr, ktx). CMake uses FetchContent to optionally pull in these dependencies replacing the old usage of submodules. This way if an option is not selected the dependencies are not checked out which significantly reduces repository checkout time and size.


CMakeLists.txt
```diff
@@ -1,17 +1,10 @@
...
-set(KTX_FEATURE_VULKAN OFF)
-set(KTX_FEATURE_GL_UPLOAD OFF)
-set(KTX_FEATURE_TOOLS OFF)
-set(KTX_FEATURE_DOC OFF)
-set(KTX_FEATURE_STATIC_LIBRARY ON)
-
-add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/third_party/glm)
-add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/third_party/tinyobjloader)
-add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/third_party/KTX-Software)
+project(gfx)
```
maoliver-amd, line 3: Adding a project reduces cmake warnings about missing project

CMakeLists.txt
```diff
@@ -1,17 +1,10 @@
...
 
 option(GFX_BUILD_EXAMPLES "Build gfx examples" ON)
+option(GFX_ENABLE_GUI "Build gfx with imgui support" OFF)
+option(GFX_ENABLE_SCENE "Build gfx with scene loading support" OFF)
```
maoliver-amd, line 7: The 2 new options. Its also possible to add extra finer grained options such as enable-ktx etc. at a later point

CMakeLists.txt
```diff
@@ -20,30 +13,27 @@ add_library(gfx STATIC gfx.cpp)
 set(GFX_DXC_PATH ${CMAKE_CURRENT_SOURCE_DIR}/third_party/dxc)
 set(GFX_PIX_PATH ${CMAKE_CURRENT_SOURCE_DIR}/third_party/WinPixEventRuntime)
 
-set(GFX_DXC_PATH ${GFX_DXC_PATH} PARENT_SCOPE)
-set(GFX_PIX_PATH ${GFX_PIX_PATH} PARENT_SCOPE)
+if(NOT PROJECT_IS_TOP_LEVEL)
+	set(GFX_DXC_PATH ${GFX_DXC_PATH} PARENT_SCOPE)
+	set(GFX_PIX_PATH ${GFX_PIX_PATH} PARENT_SCOPE)
+endif()
```
maoliver-amd, line 19: Avoid cmake warnings by only setting in parent scope when there actually is a parent scope

examples/common/fly_camera.h
```diff
@@ -23,7 +23,7 @@ SOFTWARE.
 ****************************************************************************/
 #pragma once
 
-#include "gfx_scene.h"
+#include "glm/glm.hpp"
```
maoliver-amd, line 26: Only used glm include indirectly and never GfxScene

gfx_scene.h
```diff
@@ -468,23 +468,20 @@ template<typename TYPE> bool gfxSceneSetObjectMetadata(GfxScene scene, uint64_t
 #include <map>
 #include <set>
 #include <functional>
-#pragma warning(push)
-#pragma warning(disable:4789)   // buffer will be overrun
```
maoliver-amd, line 472: Old warnings that dont appear with newer dep versions.

CMakeLists.txt
```diff
@@ -1,17 +1,10 @@
-cmake_minimum_required(VERSION 3.10.0)
+cmake_minimum_required(VERSION 3.24.0)
```
maoliver-amd, line 1: Needed to bump min cmake version as the current was already older than some of the features we were using

CMakeLists.txt
```diff
@@ -20,30 +13,27 @@ add_library(gfx STATIC gfx.cpp)
 set(GFX_DXC_PATH ${CMAKE_CURRENT_SOURCE_DIR}/third_party/dxc)
 set(GFX_PIX_PATH ${CMAKE_CURRENT_SOURCE_DIR}/third_party/WinPixEventRuntime)
 
-set(GFX_DXC_PATH ${GFX_DXC_PATH} PARENT_SCOPE)
-set(GFX_PIX_PATH ${GFX_PIX_PATH} PARENT_SCOPE)
+if(NOT PROJECT_IS_TOP_LEVEL)
+    set(GFX_DXC_PATH ${GFX_DXC_PATH} PARENT_SCOPE)
+    set(GFX_PIX_PATH ${GFX_PIX_PATH} PARENT_SCOPE)
+endif()
```
maoliver-amd, line 19: Takahiro remove this in his PR but instead I just put them in a conditional. If you run cmake on gfx alone (i.e. its the top level project) then these lines used to throw warnings about non-existent parent scope. Checking for a higher level fixes this

CMakeLists.txt
```diff
@@ -68,7 +58,113 @@ set_target_properties(WinPixEventRuntime PROPERTIES
...
+    set_target_properties(astcenc-avx2-static PROPERTIES FOLDER "third_party/ktx_deps")
+    set_target_properties(ktx_version PROPERTIES FOLDER "third_party/ktx_deps")
+    set_target_properties(fmt PROPERTIES FOLDER "third_party/ktx_deps")
+
+    if(TARGET mkvk)
+        set_target_properties(mkvk PROPERTIES FOLDER "third_party/ktx_deps")
+        set_target_properties(makedfd2vk PROPERTIES FOLDER "third_party/ktx_deps")
+        set_target_properties(makevk2dfd PROPERTIES FOLDER "third_party/ktx_deps")
+        set_target_properties(mkvkformatfiles PROPERTIES FOLDER "third_party/ktx_deps")
+    endif()
```
maoliver-amd, line 166: visual studio project organisation was moved from capsaicin to here as then it applies to all of them and then to any other project using gfx

CMakeLists.txt
gboisse: Issue here I guess is that the `gfx` examples won't build and run out of the box with such a change.
So I think it'd be best to have them on by default.

third_party/.gitignore
```diff
@@ -0,0 +1,8 @@
+/imgui/
+/stb/
+/KTX-Software/
+/Vulkan-Headers/
+/glm/
+/tinyobjloader/
+/cgltf/
+/tinyexr/
```
gboisse, line 8: Why the need to ignore these directories? Is anything getting picked up by git otherwise?

CMakeLists.txt
maoliver-amd: I did update the examples cmake as some will still build without imgui/scene. So when disabled only a subset of the examples will build

third_party/.gitignore
maoliver-amd: Since these are no longer submodules and are optionally pulled in by cmake, then when they are downloaded they go to the above locations and then git picks them up and shows and you get 1000s of new files showing up. So it pollutes the git file changes which is why they are ignored

CMakeLists.txt
maoliver-amd: OK just ran a test to double check. The issue I was trying to avoid by having the options disabled by default is that when they are enabled as default then cmake will start pulling in all the 3rd party dependencies before the user has an option to disable them. So it downloads an additional 1.18GB of dependencies automatically and even if you then disable the option first chance you get its to late as the download has finished at this point. However if the user wants to download dependencies then the first configuration only takes a couple of seconds at which point they can enable the options and then choose to pull in the dependencies without issue and have all examples working
So having the no-dependency option be usable without pulling in all the dependencies automatically on first run anyway, I was thinking is useful for creating minimal test cases etc. for sending in bug examples or for the NN work (such as the one I created last week and used this branch in)

maoliver-amd: Sorry should have been clear, with cmake GUI (emphasis on the GUI) as it does split passes whereas the cmd line does a single pass. So if you run command line and pass to disable GUI/scene from the start then it should work as we would expect. But in the GUI it picks up the default values when you fist load the project and will download all the dependencies before you get the option in the UI to tell it not to. So it depends on what use cases we wanted to support

gboisse: I was thinking of the case of mininn where say they'd already have the cmake file set up correctly at the top level, then a "new developer" who'd check out the project wouldn't get to download these dependencies even if they're enabled by default in gfx, correct?

maoliver-amd: not sure I understand.
Aslong as the cmake options GFX_ENABLE_GUI/GFX_ENABLE_SCENE  are enabled at any point then the dependencies will be downloaded. So if MiniNN pulls in gfx as a submodule with those options disabled in the parent project (providing they are set before the gfx sub project was added) then they would override any defaults in gfx and the dependencies wont be downloaded within that submodule. If a new developer checks out mininn then they would get the same thing as its being overwritten by the minnn parent project. So to get the deps to download the option must be enabled in the parent minn project (this can be done by just ticking the box in cmake GUI - this is why its way easier to enable an option later than it is to disable it as disabling requires clearing caches etc). 
Are you thinking of a new developer who arent familiar with gfx and are learning how to use it so you want them to see the full examples suite? Well either they can just create a standalone gfx clone to play with or do as I mentioned earlier and toggle the option on in the cmake GUI as this wont effect any other users of the project as its a local only change not visible to git

PR #84, opened 2023-09-15
Skinning

skkkksdkfak:


gfx.h
```diff
@@ -6280,7 +6280,7 @@ class GfxInternal
             }
             state_object_properties->Release();
         }
-        if(kernel.vertex_stride_ > 0)
+        if(!is_compute)
```
gboisse, line 6283: Any reason for reverting this change? This condition had to be changed to `if(kernel.vertex_stride_ > 0)` due to the addition of mesh shaders: https://github.com/gboisse/gfx/pull/81.

gfx.h
```diff
@@ -6305,7 +6305,7 @@ class GfxInternal
                 installed_index_buffer_ = bound_index_buffer_;
                 force_install_index_buffer_ = false;
             }
-            if(force_install_vertex_buffer_ || bound_vertex_buffer_.handle != installed_vertex_buffer_.handle)
+            if(kernel.vertex_stride_ > 0 && (force_install_vertex_buffer_ || bound_vertex_buffer_ != installed_vertex_buffer_))
```
gboisse, line 6308: Any reason behind this change?

gfx.h
skkkksdkfak: As I implemented manual vertex fetch in the vertex shader for skinned mesh, vertex_stride_ can be 0 although we still should set index buffer via IASetIndexBuffer. But, given that there are mesh shaders, I think it should be further adjusted.

gfx.h
skkkksdkfak: Because I've removed the `kernel.vertex_stride_ > 0` check from above, to allow setting index buffer below, we should ignore this branch if there are no vertex buffers for input assembler.

gfx.h
gboisse: I see.

So, we now have a proper check available to query whether a kernel uses standard graphics pipeline:
https://github.com/gboisse/gfx/blob/77ac4c63495ff0f6ad3dab8075c1c3e7beba9ef0/gfx.h#L1082

Think this check should therefore be `if(kernel.isGraphics())`, which would ensure that the IB is being set if there's one (even without a vertex buffer) and would skip these calls when using a mesh shader-based pipeline.

gboisse: btw, bit unrelated but I recently went and chased down all instances of unreleased D3D12 resources that we had and realised that this DXR-1.0 code was leaking handles every single frame (well, every dispatch to be more precise):
https://github.com/gboisse/gfx/blob/77ac4c63495ff0f6ad3dab8075c1c3e7beba9ef0/gfx.h#L6216-L6217

So FYI, `QueryInterface()` does "acquire" the handle meaning you do need to release it afterwards:
https://github.com/gboisse/gfx/blob/77ac4c63495ff0f6ad3dab8075c1c3e7beba9ef0/gfx.h#L6281

gfx_scene.h
```diff
@@ -26,6 +26,7 @@ SOFTWARE.
 
 #include "gfx.h"
 #include "glm/glm.hpp"
+#include "glm/gtc/type_ptr.hpp"
```
gboisse, line 29: Please move this include to the implementation details section so we try and keep the "public" header as clean from includes as possible.

skkkksdkfak: Done.

gfx.h
skkkksdkfak: Done.

PR #88, opened 2023-09-28
Allow instancing with different materials on same mesh.

maoliver-amd:
This moves materials from being attached to each mesh to being attached to the instance. This allows for the same mesh to be used with different materials.
The GLTF importer was updated to correctly import scenes that have this type of instancing

### DIFF
--- a/gfx_scene.h
+++ b/gfx_scene.h
@@ -1934,8 +1934,8 @@ private:
                         mesh_metadata.object_name += ".";
                         mesh_metadata.object_name += std::to_string(j);
                     }
-                    current_mesh                = mesh_ref;
-                    meshInstances[index_buffer] = mesh_ref;
+                    current_mesh                   = mesh_ref;
+                    meshInstances[position_buffer] = mesh_ref;
                 }
                 else
                 {
--- a/gfx_scene.h
+++ b/gfx_scene.h
@@ -379,8 +379,6 @@ struct GfxJoint
 
 struct GfxMesh
 {
-    GfxConstRef<GfxMaterial> material;
-
     glm::vec3 bounds_min = glm::vec3(0.0f);
     glm::vec3 bounds_max = glm::vec3(0.0f);
 
@@ -406,8 +404,9 @@ bool gfxSceneSetMeshMetadata(GfxScene scene, uint64_t mesh_handle, GfxMetadata c
 
 struct GfxInstance
 {
-    GfxConstRef<GfxMesh> mesh;
-    GfxConstRef<GfxSkin> skin;
+    GfxConstRef<GfxMesh>     mesh;
+    GfxConstRef<GfxMaterial> material;
+    GfxConstRef<GfxSkin>     skin;
 
     glm::mat4 transform = glm::mat4(1.0f);
 };
@@ -1350,8 +1349,6 @@ private:
                     mesh_metadata.object_name += ".";
                     mesh_metadata.object_name += std::to_string(mesh_id);
                 }
-                if((*it).first < materials.size())
-                    mesh_ref->material = materials[(*it).first];
                 glm::vec3 bounds_min(FLT_MAX), bounds_max(-FLT_MAX);
                 for(VertexMap::const_iterator it2 = vertices.begin(); it2 != vertices.end(); ++it2)
                 {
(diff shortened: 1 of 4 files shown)


gfx_scene.h
```diff
@@ -379,7 +377,8 @@ bool gfxSceneSetMeshMetadata(GfxScene scene, uint64_t mesh_handle, GfxMetadata c
 
 struct GfxInstance
 {
-    GfxConstRef<GfxMesh> mesh;
+    GfxConstRef<GfxMesh>     mesh;
+    GfxConstRef<GfxMaterial> material;
```
maoliver-amd, line 381: Material is now in GfxInstance instead of GfxMesh as this allows for same mesh to have multiple materials in different instances

gfx_scene.h
```diff
@@ -1597,155 +1595,169 @@ class GfxSceneInternal
             if(it != images.end()) material.ao_map = (*it).second;
             materials[&gltf_material] = material_ref;
         }
-        std::map<cgltf_mesh const *, std::vector<GfxConstRef<GfxMesh>>> meshes;
+        typedef std::pair<GfxConstRef<GfxMesh>, GfxConstRef<GfxMaterial>> instance_pair;
+        std::map<cgltf_mesh const *, std::vector<instance_pair>>          meshes;
+        std::map<cgltf_accessor const *, GfxConstRef<GfxMesh>>            meshInstances;
```
maoliver-amd, line 1600: In GLTF mesh primitives have an index into the buffer region aswell as material and attribute data. Instancing in GLTF occurs when multiple mesh primitives have the same buffer index. So `meshInstances` is introduced to track which buffer indexes (used as a gltf accessor) correspond to which mesh.
We then back up the material used for that primitive as gfx uses that in combination with the transform when making a GfxInstance so it must be stored for later

gfx_scene.h
```diff
@@ -1597,155 +1595,169 @@ class GfxSceneInternal
...
-            std::vector<GfxConstRef<GfxMesh>> &mesh_list = meshes[&gltf_mesh];
+            std::vector<instance_pair> &mesh_list = meshes[&gltf_mesh];
             for(size_t j = 0; j < gltf_mesh.primitives_count; ++j)
             {
                 cgltf_primitive const &gltf_primitive = gltf_mesh.primitives[j];
                 if(gltf_primitive.targets_count > 0) continue;   // morph targets aren't supported
                 if(gltf_primitive.type != cgltf_primitive_type_triangles) continue;    // only support triangle meshes
-                cgltf_accessor const *index_buffer = gltf_primitive.indices;
-                cgltf_attribute *gltf_primitive_attributes_end = gltf_primitive.attributes + gltf_primitive.attributes_count;
-                cgltf_attribute const *it = std::find_if(gltf_primitive.attributes, gltf_primitive_attributes_end,
```
maoliver-amd, line 1612: All this and below is the same just indented

gboisse:
Nice change, any reason to keep it as draft?

maoliver-amd:
> Nice change, any reason to keep it as draft?

Merge conflicts, it was easier to back out and then start again

gfx_scene.h
```diff
@@ -1738,204 +1737,218 @@ class GfxSceneInternal
...
-                };
-                bool failed_unpack = !unpack_sparse(position_buffer, sparse_position_buffer);
-                failed_unpack      = failed_unpack || !unpack_sparse(normal_buffer, sparse_normal_buffer);
-                failed_unpack      = failed_unpack || !unpack_sparse(uv_buffer, sparse_uv_buffer);
-                failed_unpack      = failed_unpack || !unpack_sparse(joints_buffer, sparse_joints_buffer);
-                failed_unpack      = failed_unpack || !unpack_sparse(weights_buffer, sparse_weights_buffer);
-                failed_unpack      = failed_unpack || !unpack_sparse(index_buffer, sparse_index_buffer);
-                if(failed_unpack)
+                std::map<cgltf_accessor const *, GfxConstRef<GfxMesh>>::const_iterator mesh_it = meshInstances.find(position_buffer);//Note: assumes primitives with same position buffer will always have same attributes
+                if(mesh_it == meshInstances.end())
```
maoliver-amd, line 1759: This uses the position index as the mesh primitive key (some scenes have meshes with same indices but different positions so use positions). In the GLTF spec it is in theory possible for multiple meshes to have the same position index but differences in the other attributes (i.e. different vertex normals) but this should never actually happen

PR #91, opened 2023-10-24
Use DXGI_PRESENT_ALLOW_TEARING when not using vsync.

maoliver-amd:
`DXGI_PRESENT_ALLOW_TEARING` flag is required to support variable refresh rate monitors. It is also useful as it smooths out frame times when not using vsync (looks like a pulse wave otherwise).

[https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/variable-refresh-rate-displays](https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/variable-refresh-rate-displays)
[https://learn.microsoft.com/en-us/windows/win32/api/dxgi/ne-dxgi-dxgi_swap_chain_flag](https://learn.microsoft.com/en-us/windows/win32/api/dxgi/ne-dxgi-dxgi_swap_chain_flag)

### DIFF
--- a/gfx.h
+++ b/gfx.h
@@ -1424,8 +1424,9 @@ public:
         swap_chain_desc.Format           = DXGI_FORMAT_R8G8B8A8_UNORM;
         swap_chain_desc.BufferCount      = max_frames_in_flight_;
         swap_chain_desc.BufferUsage      = DXGI_USAGE_RENDER_TARGET_OUTPUT;
-        swap_chain_desc.SwapEffect       = DXGI_SWAP_EFFECT_FLIP_SEQUENTIAL;
+        swap_chain_desc.SwapEffect       = DXGI_SWAP_EFFECT_FLIP_DISCARD;
         swap_chain_desc.SampleDesc.Count = 1;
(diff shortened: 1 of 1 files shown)


gfx.h
```diff
@@ -1424,8 +1424,9 @@ class GfxInternal
         swap_chain_desc.Format           = DXGI_FORMAT_R8G8B8A8_UNORM;
         swap_chain_desc.BufferCount      = max_frames_in_flight_;
         swap_chain_desc.BufferUsage      = DXGI_USAGE_RENDER_TARGET_OUTPUT;
-        swap_chain_desc.SwapEffect       = DXGI_SWAP_EFFECT_FLIP_SEQUENTIAL;
+        swap_chain_desc.SwapEffect       = DXGI_SWAP_EFFECT_FLIP_DISCARD;
```
maoliver-amd, line 1427: DXGI_SWAP_EFFECT_FLIP_DISCARD is supposedly slightly faster. I couldnt find any reason why we were using DXGI_SWAP_EFFECT_FLIP_SEQUENTIAL

PR #92, opened 2023-10-25
Add debug message callback.

maoliver-amd:
As it says on the tin
Errors or critical failures will trigger a debug break point while warnings will just print out the message. Any lower level information is ignored.
The message callback isnt unregistered as its used for the life time of the device anyway

### DIFF
--- a/gfx.h
+++ b/gfx.h
@@ -1411,14 +1411,9 @@ public:
                                                 D3D12_MESSAGE_SEVERITY Severity, D3D12_MESSAGE_ID ID,
                                                 LPCSTR pDescription, void *pContext) {
                     if(Severity <= D3D12_MESSAGE_SEVERITY_ERROR)
-                    {
-                        GFX_PRINTLN("D3D12 Error: %s", pDescription);
-                        GFX_ASSERT(false);
-                    }
+                        GFX_ASSERTMSG(0, "D3D12 Error: %s", pDescription);
                     else if(Severity == D3D12_MESSAGE_SEVERITY_WARNING)
-                    {
                         GFX_PRINTLN("D3D12 Warning: %s", pDescription);
-                    }
                 };
                 DWORD cookie;
                 debugCallback->RegisterMessageCallback(
--- a/gfx.h
+++ b/gfx.h
@@ -1402,6 +1402,31 @@ public:
         device_->QueryInterface(IID_PPV_ARGS(&mesh_device_));
         SetDebugName(device_, "gfx_Device");
 
+        if((flags & kGfxCreateContextFlag_EnableDebugLayer) != 0)
+        {
+            ID3D12InfoQueue1 *debugCallback;
+            if(SUCCEEDED(device_->QueryInterface(IID_PPV_ARGS(&debugCallback))))
+            {
+                D3D12MessageFunc callback = [](D3D12_MESSAGE_CATEGORY  Category,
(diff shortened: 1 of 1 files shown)


gfx.h
```diff
@@ -1402,6 +1402,24 @@ class GfxInternal
         device_->QueryInterface(IID_PPV_ARGS(&mesh_device_));
         SetDebugName(device_, "gfx_Device");
 
+        if((flags & kGfxCreateContextFlag_EnableDebugLayer) != 0)
+        {
+            ID3D12InfoQueue1 *debugCallback;
+            device_->QueryInterface(IID_PPV_ARGS(&debugCallback));
```
gboisse, line 1408: Do we not need a `debugCallback->Release()` once we're done with this?

gfx.h
```diff
@@ -1402,6 +1402,24 @@ class GfxInternal
...
+                                            D3D12_MESSAGE_ID ID, LPCSTR pDescription, void *pContext) {
+                if(Severity <= D3D12_MESSAGE_SEVERITY_ERROR)
+                {
+                    GFX_PRINTLN("D3D12 Error: %s", pDescription);
+                    GFX_ASSERT(false);
+                } else if(Severity == D3D12_MESSAGE_SEVERITY_WARNING) {
+                    GFX_PRINTLN("D3D12 Warning: %s", pDescription);
+                }
+            };
+            debugCallback->RegisterMessageCallback(callback, D3D12_MESSAGE_CALLBACK_FLAG_NONE, this, &cookie);
```
gboisse, line 1420: `debugCallback` can be null here (for instance, by disabling Developer Mode in Windows Developer Settings I believe? or if the Graphics Tools optional feature isn't installed) so we should have a pointer check to avoid crashing.

gfx.h
```diff
@@ -1402,6 +1402,24 @@ class GfxInternal
...
+        {
+            ID3D12InfoQueue1 *debugCallback;
+            device_->QueryInterface(IID_PPV_ARGS(&debugCallback));
+            DWORD cookie;
+            D3D12MessageFunc callback = [](D3D12_MESSAGE_CATEGORY Category, D3D12_MESSAGE_SEVERITY Severity,
+                                            D3D12_MESSAGE_ID ID, LPCSTR pDescription, void *pContext) {
+                if(Severity <= D3D12_MESSAGE_SEVERITY_ERROR)
+                {
+                    GFX_PRINTLN("D3D12 Error: %s", pDescription);
+                    GFX_ASSERT(false);
```
gboisse, line 1415: You could use `GFX_ASSERTMSG(0, "D3D12 Error: %s", pDescription);`, which would do the same thing but in a single line.

gfx.h
maoliver-amd: yes, ill add that

gfx.h
maoliver-amd: fixed

gfx.h
maoliver-amd: I actually left it as 2 lines so that the debug message was already printed by the time the assert was triggered. The messages are rather long so its quicker and easier to view them once printed

gboisse: `GFX_ASSERTMSG()` does print out the message prior to the breakpoint, so that should be the same behaviour.

maoliver-amd: fair enough, changed

PR #93, opened 2023-11-06
Add function to create auto resized texture with mips.

maoliver-amd:
Allows creating auto resized textures but with mip level creation.

### DIFF
--- a/gfx.h
+++ b/gfx.h
@@ -8634,10 +8634,16 @@ private:
             D3D12MA::Allocation *allocation = nullptr;
             D3D12_RESOURCE_DESC
             resource_desc        = texture.resource_->GetDesc();
+            if(resource_desc.MipLevels > 1)
+            {
+                uint32_t max_mips = gfxCalculateMipCount(window_width, window_height);
+                if(resource_desc.MipLevels == (UINT16)gfxCalculateMipCount((uint32_t)resource_desc.Width, (uint32_t)resource_desc.Height))
+                    resource_desc.MipLevels = max_mips;
+                else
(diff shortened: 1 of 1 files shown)


gfx.h
```diff
@@ -8627,6 +8634,14 @@ class GfxInternal
...
             D3D12_RESOURCE_DESC
             resource_desc        = texture.resource_->GetDesc();
+            if(resource_desc.MipLevels > 1)
+            {
+                uint32_t max_mips = gfxCalculateMipCount(window_width, window_height);
+                if(resource_desc.MipLevels == (UINT16)gfxCalculateMipCount((uint32_t)resource_desc.Width, (uint32_t)resource_desc.Height))
+                    resource_desc.MipLevels = max_mips;
+                else
+                    resource_desc.MipLevels = GFX_MIN((uint32_t)resource_desc.MipLevels, max_mips);
+            }
```
maoliver-amd, line 8644: The idea here is that if the user specifies a mip count below the default for the current resolution then always try and retain that. However if using auto mip levels then it should auto resize the mip count based on resolution change (auto mip levels is detected by checking if current set value equals calculated levels for that resolution).

Technically its possible that the user might specify a low fixed mip count then the screen gets resized such that the low mip count now matches the auto mip count in which on screen resize up again the mips will start getting auto resized instead of staying at the low value. However this cant be supported without adding extra stored values and is a rather unusual case that is currently just not supported.

maoliver-amd: Alternatively, an additional flag could be added to the internal texture to indicate auto mip map creation and then add an additional function to create a auto sized exture with auto mip maps (e.g. createTexture2DMips)

PR #94, opened 2023-11-07
Update winpix to 1.0.231030001.

maoliver-amd:
Updates winpix to 1.0.231030001 and also updates the AMD pix3 markers from RDTS 2023-12-04-1282.


gboisse:
Thanks for the update, I assume this hasn't lost any of the RGP integration + fixes?
i.e., do named sections still show up in RGP + no D3D12 leak is reported upon termination in `Debug` builds?

maoliver-amd:
I havnt actually tested this in RGP, just in PIX. Apparently they changed alot to improve performance and made some breaking changes to the pix data format. So I have mainly just tested that.
Leaks are reported the same as they were before (apparently we had 1) so that seems to work as before

maoliver-amd:
The latest RDTS has been released and includes updated headers and install documentation for the latest version of pix. Ive updated this PR so everything should be working now.

gboisse:
Could confirm it's working well, I can see the event names in an RGP capture with latest version of RDTS.

PR #99, opened 2024-01-05
Implement morphing support

picul:


gfx_scene.h
```diff
@@ -1690,177 +1730,164 @@ class GfxSceneInternal
...
-                            }
-                        }
-                        return true;
-                    };
-                    bool failed_unpack = !unpack_sparse(position_buffer, sparse_position_buffer);
-                    failed_unpack      = failed_unpack || !unpack_sparse(normal_buffer, sparse_normal_buffer);
-                    failed_unpack      = failed_unpack || !unpack_sparse(uv_buffer, sparse_uv_buffer);
-                    failed_unpack      = failed_unpack || !unpack_sparse(joints_buffer, sparse_joints_buffer);
-                    failed_unpack      = failed_unpack || !unpack_sparse(weights_buffer, sparse_weights_buffer);
-                    failed_unpack      = failed_unpack || !unpack_sparse(index_buffer, sparse_index_buffer);
```
gboisse, line 1751: Any reason for removing these `unpack_sparse()` calls? They don't seem to appear anywhere in the new code.

gboisse (commented):
Seems ok to me overall, although I am a bit concerned about the removal of the sparse buffer support, any reason for this?
In general, I think we should avoid removing previously supported features...
Adding @maoliver-amd as a reviewer too.

gfx_scene.h
maoliver-amd: `UnpackAccessor` just below does the same thing. Although it now always uses the unpack floats function for sparse and non-sparse accessors. It wasnt done that way previously and I have no idea why, I think we had a sparse/non-sparse code path for performance reasons but I cant remeber. It would be interesting to know what performance difference it makes having only a single code path

maoliver-amd (approved):
I didnt see I had been pinged for review (got lost in the holiday emails apparently).

Anyway theres a fair bit here to go through but ive gone through everything and it looks fine. Its been a while since I wrote alot of this (and I admit I had to look up the cgltf docs a few times) but everything appears good. Ultimately the proof is in functionality and we dont have alot of test scenes in capsaicin at the moment. But assuming all scenes load correctly then the code itself looks ok.

gfx_scene.h
picul: Well I added `UnpackAccessor` in order to fix this comment. And I decided to always do unpacking since it's just much simpler in terms of implementation and readability. I guess it might have negative impact on performance, but it's surely not too large and is limited to model loading time. Unfortunately I just don't have large models with sparse accessors to test it.

PR #102, opened 2024-02-08
Add function to copy a texture to buffer

maoliver-amd:
This adds a new function `gfxCommandCopyTextureToBuffer` that can be used to copy a texture into a host readable buffer.

Tested to dump AOV to disk as well as round trip from bufeer to texture and then back to buffer with input and output contents identical for images of power-of-2 and non-power-of-2 and with and without mip chains.

I also added to this PR a new path within `gfxCommandCopyTexture` that will handle copying between textures with different mip levels. This allows (and was tested) copying a texture into the first mip level of another texture

### DIFF
--- a/gfx.h
+++ b/gfx.h
@@ -3428,55 +3428,18 @@ public:
         submitPipelineBarriers();   // transition our resources if needed
         for(uint32_t mip_level = 0; mip_level < src.mip_levels; ++mip_level)
         {
-            Buffer *texture_copy_buffer = nullptr;
             if(buffer_offset >= dst.size)
                 break;  // further mips aren't available
             uint64_t const buffer_row_pitch = row_sizes[mip_level];
-            uint64_t const texture_row_pitch = subresource_footprints[mip_level].Footprint.RowPitch;
             uint64_t const buffer_size = (uint64_t)num_rows[mip_level] * buffer_row_pitch;
             if(buffer_offset + buffer_size > dst.size)
                 return GFX_SET_ERROR(kGfxResult_InvalidOperation, "Cannot copy from mip level %u to buffer object with insufficient storage", mip_level);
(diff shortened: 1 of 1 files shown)


gfx.h
```diff
@@ -3344,54 +3345,80 @@ class GfxInternal
...
-                transitionResource(*texture_upload_buffer, D3D12_RESOURCE_STATE_COPY_DEST);
-                submitPipelineBarriers();   // transition our resources if needed
-                for(uint32_t i = 0; i < num_rows[mip_level]; ++i)
-                    command_list_->CopyBufferRegion(texture_upload_buffer->resource_, i * texture_row_pitch, gfx_buffer.resource_, i * buffer_row_pitch + buffer_offset, buffer_row_pitch);
-            }
-            Buffer &gfx_buffer = buffers_[src]; SetObjectName(gfx_buffer, src.name);
-            if(texture_upload_buffer != nullptr) transitionResource(*texture_upload_buffer, D3D12_RESOURCE_STATE_COPY_SOURCE);
-            else if(src.cpu_access == kGfxCpuAccess_None) transitionResource(gfx_buffer, D3D12_RESOURCE_STATE_COPY_SOURCE);
-            transitionResource(gfx_texture, D3D12_RESOURCE_STATE_COPY_DEST);
-            submitPipelineBarriers();   // transition our resources if needed
```
maoliver-amd, line 3384: I also noticed that the existing buffer to texture functions does a series of smaller CopyBufferRegion calls when buffer and texture pitch dont match. However I found that by just setting the Footprint.RowPitch in the CopyTextureRegion function makes this unnecessary. I checked this in PIX using input buffers containing mip chains that trigger this code path and it works identically after this change. It is also still round trip identical when combined with the textureToBuffer function.
This change is in its own commit so can be skipped/reverted if not wanted.

gboisse (approved):

PR #103, opened 2024-02-15
Dont unbind index/vertex buffers.

maoliver-amd:
Existing code was unbinding any index/vertex buffers if they were found in the additional parameter list. This causes draw errors when a input parameter is also the bound index/vertex buffer as previous behaviour was unbinding them just before they are needed in a draw call. The unbinding doesnt seem to be needed (comment was about debug layers but testing on both NVIDIA and AMD shows no debug warnings) so was removed.

I also fixed an error in the animation code while I was at it.

### DIFF
--- a/gfx.h
+++ b/gfx.h
@@ -8721,12 +8721,6 @@ private:
                 submitPipelineBarriers();
                 break;  // break barrier batches to avoid debug layer warnings
             }
-        if(*buffer.resource_state_ == D3D12_RESOURCE_STATE_INDEX_BUFFER &&  // unbind if active index buffer to prevent debug layer errors
-           buffer_handles_.has_handle(bound_index_buffer_.handle) && buffers_[bound_index_buffer_].resource_ == buffer.resource_)
-            command_list_->IASetIndexBuffer(nullptr);
-        if(*buffer.resource_state_ == D3D12_RESOURCE_STATE_VERTEX_AND_CONSTANT_BUFFER &&    // unbind if active vertex buffer to prevent debug layer errors
-           buffer_handles_.has_handle(bound_vertex_buffer_.handle) && buffers_[bound_vertex_buffer_].resource_ == buffer.resource_)
-            command_list_->IASetVertexBuffers(0, 1, nullptr);
         D3D12_RESOURCE_BARRIER resource_barrier = {};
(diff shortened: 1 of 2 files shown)


gfx_scene.h
```diff
@@ -1152,7 +1152,7 @@ class GfxSceneInternal
                     animation_channel.values_[4 * next_index + 1], animation_channel.values_[4 * next_index + 2]);
                 animated_node->rotate_ = glm::slerp(previous, next, interpolate);
             }
-            if(animation_channel.type_ == kGltfAnimationChannelType_Scale)
+            else if(animation_channel.type_ == kGltfAnimationChannelType_Scale)
```
maoliver-amd, line 1155: Without this 'else if' then the code will execute the proceeding 'else' branch which is specific to animations of type `kGltfAnimationChannelType_Weights`. However without this fixe it was being executed for all animation types other the 'scale' which was just generating erroneous data.

gfx.h
```diff
@@ -8721,12 +8721,6 @@ class GfxInternal
                 submitPipelineBarriers();
                 break;  // break barrier batches to avoid debug layer warnings
             }
-        if(*buffer.resource_state_ == D3D12_RESOURCE_STATE_INDEX_BUFFER &&  // unbind if active index buffer to prevent debug layer errors
-           buffer_handles_.has_handle(bound_index_buffer_.handle) && buffers_[bound_index_buffer_].resource_ == buffer.resource_)
-            command_list_->IASetIndexBuffer(nullptr);
-        if(*buffer.resource_state_ == D3D12_RESOURCE_STATE_VERTEX_AND_CONSTANT_BUFFER &&    // unbind if active vertex buffer to prevent debug layer errors
-           buffer_handles_.has_handle(bound_vertex_buffer_.handle) && buffers_[bound_vertex_buffer_].resource_ == buffer.resource_)
-            command_list_->IASetVertexBuffers(0, 1, nullptr);
```
maoliver-amd, line 8729: This was unbinding vertex/index buffers if they were also used as a program parameter. Not having a bound vertex buffer generally breaks rendering so it was removed.
Alternatively we could move binding vertex/index buffers to the program/kernel level within gfx instead of globally setting them as this way we would know if a bound buffer is required or not and could default to unbinding for any kernel launch that doesnt need it.

gboisse (approved):

PR #105, opened 2024-02-21
Add functions to resize existing textures.

maoliver-amd:
Does what it says on the title.
This makes it easier to resize textures as it automatically handles copying over texture name and other parameters (technically doesnt copy them as it just reuses them). This also fixes an issue whereby the clear value is not user accessible so theres no easy to recreate a texture using the set clear value. With these new functions that is now fixed as the clear value is copied over from existing texture when resizing.


gfx.h
```diff
@@ -2226,6 +2230,237 @@ class GfxInternal
...
+        collect(gfx_texture); // release previous texture
+        uint64_t handle = texture_handles_.allocate_handle();
+        Texture &new_texture = textures_.insert(handle & 0xFFFFFFFFull);
+        memcpy(new_texture.clear_value_, gfx_texture.clear_value_, sizeof(gfx_texture.clear_value_));
+        new_texture.resource_ = resource;
+        new_texture.allocation_ = allocation;
+        new_texture.resource_state_ = resource_state;
+        textures_.erase(texture); // destroy old texture object
+        texture_handles_.free_handle(texture.handle);
+        texture.handle = handle;
```
maoliver-amd, line 2275: something somewhere seems to hold and use the handle, If reusing the same handle then you get rendering errors (even when the texture objects are identical) so this creates a new handle

maoliver-amd:
> What about making the clear value part of the `GfxTexture` object (with the corresponding getter) then have these "re-create" functions as part of the application? They wouldn't need any internal `gfx` access then.

There are currently 2 locations in gfx that work purely with the internal Texture without knowing what the corresponding public facing GfxTexture object is. Moving the clear value from internal to public would cause an issue here as there doesnt seem to be an easy way to get the GfxTexture from a Texture. So it required duplicating the clear value in both GfxTexture and internal Texture

gfx.h
```diff
@@ -116,7 +116,7 @@ TYPE *gfxBufferGetData(GfxContext context, GfxBuffer buffer)
 //! Texture resources.
 //!
 
-class GfxTexture { GFX_INTERNAL_NAMED_HANDLE(GfxTexture); uint32_t width; uint32_t height; uint32_t depth; DXGI_FORMAT format; uint32_t mip_levels; enum { kType_2D, kType_2DArray, kType_3D, kType_Cube } type; public:
+class GfxTexture { GFX_INTERNAL_NAMED_HANDLE(GfxTexture); uint32_t width; uint32_t height; uint32_t depth; DXGI_FORMAT format; uint32_t mip_levels; enum { kType_2D, kType_2DArray, kType_3D, kType_Cube } type; float clear_value_[4] = {}; public:
```
gboisse, line 119: No need for that `= {}` since the `GFX_INTERNAL_NAMED_HANDLE()` macro inserts a constructor that does `memset(this, 0, sizeof(*this));`.

maoliver-amd: done

PR #109, opened 2024-02-29
Fix Validation errors

maoliver-amd:
This PR fixes validation errors that occur when enabling `SetEnableGPUBasedValidation`.
These errors are due to kernel parameters being default transitioned to `D3D12_RESOURCE_STATE_PIXEL_SHADER_RESOURCE` even though the associated kernel was compute/non-pixel shader. Changing to `D3D12_RESOURCE_STATE_ALL_SHADER_RESOURCE` fixes this for all cases I could find, however, I went 1 step further and set the transition state based on the associated kernel (this last part is in a second commit that can be reverted if desired)

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -6029,7 +6029,7 @@ private:
             uint32_t descriptor_slot = (parameter.descriptor_slot_ != 0xFFFFFFFFu ? parameter.descriptor_slot_ :
                                                                                     dummy_descriptors_[parameter.type_]);
             if(parameter.descriptor_slot_ != 0xFFFFFFFFu)
-                initDescriptorParameter(program, invalidate_descriptors, parameter, descriptor_slot);
+                initDescriptorParameter(kernel, program, invalidate_descriptors, parameter, descriptor_slot);
             GFX_ASSERT(descriptor_slot < (descriptors_.descriptor_heap_ != nullptr ? descriptors_.descriptor_heap_->GetDesc().NumDescriptors : 0));
             if(is_compute)
                 command_list_->SetComputeRootDescriptorTable(i, descriptors_.getGPUHandle(descriptor_slot));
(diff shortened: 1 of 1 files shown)


gfx.cpp
```diff
@@ -7262,7 +7262,7 @@ class GfxInternal
                             device_->CreateShaderResourceView(nullptr, &dummy_srv_desc, descriptors_.getCPUHandle(parameter.descriptor_slot_ + j));
                             continue;   // out of bounds mip level
                         }
-                        transitionResource(gfx_texture, D3D12_RESOURCE_STATE_PIXEL_SHADER_RESOURCE);
+                        transitionResource(gfx_texture, kernel.isGraphics() && kernel.ps_reflection_ != nullptr ? D3D12_RESOURCE_STATE_ALL_SHADER_RESOURCE : D3D12_RESOURCE_STATE_NON_PIXEL_SHADER_RESOURCE);
```
maoliver-amd, line 7265: This also fixes a bug in gfx whereby barriers arent functioning correctly when you have a compute shader called in a loop that ping-pongs inputs betwen `Texture2D` and `RWTexture2D` (such as a kernel theat creates mip maps with each kernel call setting the Texture2D input to equal the RWTexture2D output from the previous call). Previously this type of execution failed and required setting the input parameter as a `RWTexture2D` as a workaround

gfx.cpp
```diff
@@ -7134,7 +7134,7 @@ class GfxInternal
                         Buffer &gfx_buffer = buffers_[buffer];
                         SetObjectName(gfx_buffer, buffer.name);
                         if(buffer.cpu_access == kGfxCpuAccess_None)
-                            transitionResource(gfx_buffer, D3D12_RESOURCE_STATE_PIXEL_SHADER_RESOURCE);
+                            transitionResource(gfx_buffer, kernel.isGraphics() && kernel.ps_reflection_ != nullptr ? D3D12_RESOURCE_STATE_ALL_SHADER_RESOURCE : D3D12_RESOURCE_STATE_NON_PIXEL_SHADER_RESOURCE);
```
gboisse, line 7137: Could this check be wrapped up in a helper function?
Maybe like `transitionResource(gfx_buffer, GetShaderResourceState(kernel));` or something along the lines?

maoliver-amd: done

gfx.cpp
```diff
@@ -4825,6 +4825,14 @@ class GfxInternal
...
     }
 
+    static inline D3D12_RESOURCE_STATES GetShaderVisibleResourceState(
+        Kernel const &kernel, Kernel::Parameter &parameter)
+    {
+        return kernel.isGraphics() && kernel.ps_reflection_ != nullptr
+                 ? D3D12_RESOURCE_STATE_ALL_SHADER_RESOURCE
+                 : D3D12_RESOURCE_STATE_NON_PIXEL_SHADER_RESOURCE;
+    }
+
```
gboisse, line 4835: `parameter` seems unused, let's remove it and this is good to merge.

gboisse (approved):

PR #113, opened 2024-03-08
Add dds texture loading support.

maoliver-amd:
Added support for loading dds texture files.
I couldnt find a decent thrid party lib that didnt have some sort of issues so I ended up writing some dds code from scratch. Ive tested it with all the dds files that I have and it works correctly (checked with different channel counts, compression and mips)
I also extended the GLTF loader so that it can pick up dds files wrapped in the "MSFT_texture_dds" official extension (although we can load dds files that are added without using th official extension aswell)


CMakeLists.txt
```diff
@@ -99,6 +99,8 @@ if(GFX_ENABLE_GUI)
     target_compile_definitions(gfx PUBLIC GFX_ENABLE_GUI)
 endif()
 if(GFX_ENABLE_SCENE)
+    option(GFX_ENABLE_SCENE_KTX "Build gfx with ktx image loading support" ON)
```
maoliver-amd, line 102: Also added an option so that ktx support is optional as the ktx libs are rather heavy and with dds they may not be wanted (kept them default enabled for backwards compat)

gfx_scene.cpp
```diff
@@ -1016,7 +1310,7 @@ class GfxSceneInternal
                 if(gfxSceneImport(scene, image_file.c_str()) != kGfxResult_NoError)
                     continue; // unable to load image file
                 image_ref          = gfxSceneFindObjectByAssetFile<GfxImage>(scene, image_file.c_str());
-                images[gltf_image] = image_ref;
+                images[gltf_texture.image] = image_ref;
```
maoliver-amd, line 1313: Theres a slight change as now textures are looked up based on the base image pointer. This simplifies the code

PR #114, opened 2024-03-08
Additional fixes for compilation code warnings.

maoliver-amd:
This fixes all the warnings/errors that msvc throws when gfx is opened as a cmake project directly


gfx_core.h
```diff
@@ -249,6 +249,16 @@ static inline GfxResult GFX_SET_ERROR_IMPL(GfxResult result, char const *file_na
...
 
+static inline void* gfxMalloc(size_t size)
+{
+#ifdef _MSC_VER
+    _Notnull_ /* Silence msvc warnings about not checking allocation failure */
+#endif
+    void *ret = malloc(size);
+    GFX_ASSERT(ret != nullptr); // out of memory
+    return ret;
+}
```
maoliver-amd, line 260: To help cover allocation errors, I created this gfx malloc function. This allows for allocation failures to all be checked in a single location and to silence compiler warnings in one location.

gfx.cpp
```diff
@@ -3642,12 +3633,12 @@ class GfxInternal
...
-        GfxBuffer const scan_level_2_args_buffer = createBufferRange(texture_upload_buffer_, (num_groups_level_1 + num_groups_level_2 + 6) << 2, 4 << 2);
+        GfxBuffer const scan_level_1_buffer = createBufferRange(texture_upload_buffer_, 0, (uint64_t)num_groups_level_1 << 2);
+        GfxBuffer const scan_level_2_buffer = createBufferRange(texture_upload_buffer_, (uint64_t)num_groups_level_1 << 2, (uint64_t)num_groups_level_2 << 2);
+        GfxBuffer const num_groups_level_1_buffer = createBufferRange(texture_upload_buffer_, ((uint64_t)num_groups_level_1 + num_groups_level_2) << 2, 4);
+        GfxBuffer const num_groups_level_2_buffer = createBufferRange(texture_upload_buffer_, ((uint64_t)num_groups_level_1 + num_groups_level_2 + 1) << 2, 4);
+        GfxBuffer const scan_level_1_args_buffer = createBufferRange(texture_upload_buffer_, ((uint64_t)num_groups_level_1 + num_groups_level_2 + 2) << 2, 4 << 2);
+        GfxBuffer const scan_level_2_args_buffer = createBufferRange(texture_upload_buffer_, ((uint64_t)num_groups_level_1 + num_groups_level_2 + 6) << 2, 4 << 2);
```
maoliver-amd, line 3641: There were a bunch of buffer overflow warnings, which are easy fixed by casting up to uint64 during the operations instead of after

gfx.cpp
```diff
@@ -5772,8 +5755,9 @@ class GfxInternal
         for(size_t i = 0; i < kernel.subobjects_.size(); ++i)
             max_export_length = GFX_MAX(max_export_length, strlen(kernel.subobjects_[i].c_str()));
         max_export_length += 1;
-        char *lib_export = (char *)alloca(max_export_length);
-        WCHAR *wexport = (WCHAR *)alloca(max_export_length << 1);
+        char lib_export[1024];
+        WCHAR wexport[1024];
+        GFX_ASSERT(max_export_length < 1024);
```
maoliver-amd, line 5760: Apparently alloca is not well liked by the compiler, Most other places in gfx use a fixed size array instead of alloca so I changed to that instead

CMakeLists.txt
```diff
@@ -41,6 +41,12 @@ if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
 elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
     target_compile_options(gfx PRIVATE $<$<COMPILE_LANGUAGE:CXX>:/W4 /WX /experimental:external /external:anglebrackets /external:W0 /analyze:external->)
     target_compile_options(gfx PRIVATE $<$<COMPILE_LANGUAGE:CXX>:/wd4324>) # structure was padded due to alignment specifier
+    target_compile_options(gfx PRIVATE $<$<COMPILE_LANGUAGE:CXX>:/wd6387>) # param could be 0
+    target_compile_options(gfx PRIVATE $<$<COMPILE_LANGUAGE:CXX>:/wd6262>) # Function uses alot of stack
+    target_compile_options(gfx PRIVATE $<$<AND:$<COMPILE_LANGUAGE:CXX>,$<CONFIG:Debug>>:/wd6011>) # Dereferencing NULL pointer
+    target_compile_options(gfx PRIVATE $<$<AND:$<COMPILE_LANGUAGE:CXX>,$<CONFIG:Debug>>:/wd28183>) # Variable could be '0'
+    target_compile_options(gfx PRIVATE $<$<AND:$<COMPILE_LANGUAGE:CXX>,$<CONFIG:Debug>>:/wd6385>) # Reading invalid data
+    target_compile_options(gfx PRIVATE $<$<AND:$<COMPILE_LANGUAGE:CXX>,$<CONFIG:Debug>>:/wd6386>) # Buffer overrun while writing
```
maoliver-amd, line 49: Unfortunately the gfx aseert macro causes errors to be thrown that otherwise wouldnt have as such Ive bypassed them on debug builds

gfx_imgui.cpp
```diff
@@ -74,7 +83,7 @@ class GfxImGuiInternal
             float const font_size = (font_configs != nullptr && font_configs[i].SizePixels > 0.0f)
                                 ? font_configs[i].SizePixels
                                 : 16.0f;
-            io.Fonts->AddFontFromFileTTF(font_filenames[i], font_size, &font_configs[i]);
+            io.Fonts->AddFontFromFileTTF(font_filenames[i], font_size, font_configs != nullptr ? &font_configs[i] : nullptr);
```
maoliver-amd, line 86: OK this was a legit error

gfx.cpp
```diff
@@ -6349,7 +6332,7 @@ class GfxInternal
         D3D12_GPU_VIRTUAL_ADDRESS gpu_addr = {};
         Buffer *constant_buffer = buffers_.at(constant_buffer_pool_[fence_index_]);
         uint64_t constant_buffer_size = (constant_buffer != nullptr ? constant_buffer->resource_->GetDesc().Width : 0);
-        if(constant_buffer_pool_cursors_[fence_index_] * 256 + data_size > constant_buffer_size)
+        if(constant_buffer_pool_cursors_[fence_index_] * 256 + data_size > constant_buffer_size || constant_buffer == nullptr)
```
gboisse, line 6335: I suppose this condition is only here to satisfy the warnings, right?

gfx.cpp
```diff
@@ -6510,12 +6493,10 @@ class GfxInternal
             did_type_str = "uint3";
             break;
         case GfxTexture::kType_3D:
+        default:
```
gboisse, line 6496: Any reason for moving the `default` here?

It's useful to assert on unsupported switch/case values, like when you add in a new enum entry and you can then follow the asserts for the various switch/case statements that have to be updated accordingly.

gfx.cpp
```diff
@@ -6532,12 +6513,10 @@ class GfxInternal
             select_string = "float3 select(bool3 a, float3 b, float3 c){return a ? b : c;}";
             break;
         case 4:
+        default:
```
gboisse, line 6516: Idem as previous comment.

CMakeLists.txt
gboisse: Alternatively I suppose we could include some "#pragma warning push/pop" at the start and end of the `GFX_ASSERT` macro.
But I'm ok with this CMake-based approach too.

gfx.cpp
gboisse: Only issue would be running out of memory, but at least there's an assert to catch this.

Although since this is a heavy call anyway (i.e., creating a PSO) we may as well use `gfxMalloc()` here (which you wouldn't want to call on a "hot path" kind of function).

gfx.cpp
```diff
@@ -8200,8 +8179,9 @@ class GfxInternal
                 for(size_t i = 0; i < kernel.exports_.size(); ++i)
                     max_export_length = GFX_MAX(max_export_length, strlen(kernel.exports_[i].c_str()));
                 max_export_length += 1;
-                char *lib_export = (char *)alloca(max_export_length);
-                WCHAR *wexport = (WCHAR *)alloca(max_export_length << 1);
+                char lib_export[1024];
+                WCHAR wexport[1024];
+                GFX_ASSERT(max_export_length < 1024);
```
gboisse, line 8184: Same statement than before, maybe we should just use `gfxMalloc()` here?
This might actually be related to this issue now that I think of it 🙂 https://github.com/Radeon-Pro/Capsaicin/issues/567

gboisse:
Looks good overall, but please look at the various comments I've left on the PR, thanks!

CMakeLists.txt
maoliver-amd: Unfortunately the errors arent actually in the assert macro but instead because of the presence of checks for whatever is in the assert statement, which causes the analysis to function differently and throw an error it wouldnt otherwise throw.
So as a simple case: if we have a pointer that is flagged as not null (using the gfxMalloc workaround) then there shouldnt be any errors about de-referencing null pointers however if we put a `GFX_ASSERT(pointer != NULL)` somewhere then the analysis will now assume that the pointer can in fact now be null (as otherwise why would you be checking for it) and throw an error. We could workaround this by having assert actually terminate on error as then we would actually be handling the error in some way, wheras currently we just print (and only if a debugger is attached) and continue

gfx.cpp
maoliver-amd: It was throwing a bunch of errors about uninitialised variables etc. when taking the default path, so I made default also the 3D path. Again this could be fixed by having assert terminate, or use static assert instead. std::unreachable is only available in c++23 but we could create our own macro implementation to keep the compiler happy

gfx.cpp
maoliver-amd: yep

gfx.cpp
```diff
@@ -7773,68 +7764,73 @@ class GfxInternal
...
                     cbv_desc.SizeInBytes = GFX_ALIGN(parameter.variable_size_, 256);
                 }
-                else if(parameter.parameter_->type_ == Program::Parameter::kType_Constants)
-                {
-                    cbv_desc.BufferLocation = allocateConstantMemory(parameter.parameter_->data_size_, data);
-                    if(data != nullptr) memcpy(data, parameter.parameter_->data_.constants_, parameter.parameter_->data_size_);
-                    cbv_desc.SizeInBytes = GFX_ALIGN(parameter.parameter_->data_size_, 256);
-                }
-                else if(parameter.parameter_->type_ == Program::Parameter::kType_Buffer)
+                else
```
maoliver-amd, line 7767: The diff from this point down looks ugly but all that happened was code was indented

gfx.cpp
maoliver-amd: Changed to using heap allocated vector (for auto handled free). Modern OSs lazily allocate memory anyway so an out of memory error is never thrown on allocation (unless you exhausted the 64b virtual space) so having to check it doesnt really achieve much. But std::vector with exceptions disabled will just call std::terminate on memory allocation failure so it handles well enough for our needs

gfx.cpp
maoliver-amd: Fixed now that assert terminates

PR #115, opened 2024-03-12
Adds the DirectX agility SDK as a direct depenency.

maoliver-amd:
This adds the agility SDK as a dependency project and thus avoids potential issues with a user not having a new enough version of the the direct3d headers installed


CMakeLists.txt
```diff
@@ -30,11 +32,12 @@ target_compile_options(gfx PRIVATE /bigobj)
 
 target_include_directories(gfx PRIVATE ${GFX_DXC_PATH})
 target_include_directories(gfx PRIVATE ${GFX_PIX_PATH}/include)
+target_include_directories(gfx PUBLIC ${GFX_AGILITY_PATH}/include)
 target_include_directories(gfx PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
 target_include_directories(gfx PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/third_party/D3D12MemoryAllocator/include)
 target_include_directories(gfx PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/third_party/D3D12MemoryAllocator/src)
 
-target_compile_features(gfx PUBLIC cxx_std_17)
+target_compile_features(gfx PUBLIC cxx_std_20)
```
maoliver-amd, line 40: Agility SDK requires compiling with c++20

CMakeLists.txt
```diff
@@ -65,6 +68,17 @@ if(CMAKE_SYSTEM_PROCESSOR MATCHES "(x86)|(X86)|(amd64)|(AMD64)")
...
+add_library(D3D12Core SHARED IMPORTED)
+set_target_properties(D3D12Core PROPERTIES
+    IMPORTED_LOCATION ${GFX_AGILITY_PATH}/lib/D3D12Core.dll
+    IMPORTED_IMPLIB ${GFX_AGILITY_PATH}/lib/dummy.lib
+)
+add_library(d3d12SDKLayers SHARED IMPORTED)
+set_target_properties(d3d12SDKLayers PROPERTIES
+    IMPORTED_LOCATION ${GFX_AGILITY_PATH}/lib/d3d12SDKLayers.dll
+    IMPORTED_IMPLIB ${GFX_AGILITY_PATH}/lib/dummy.lib
+)
```
maoliver-amd, line 80: Added the agility SK libs as cmake dependencies, this allows for importing projects to correctly install them as needed

examples/00-hellotriangle/CMakeLists.txt
```diff
@@ -16,50 +16,7 @@ source_group("Media Files\\Shaders" FILES ${SHADER_FILES})
...
-    COMMAND ${CMAKE_COMMAND} -E copy
-            ${GFX_PIX_PATH}/bin/x64/WinPixEventRuntime.dll
-            ${CMAKE_CURRENT_BINARY_DIR}/Release/WinPixEventRuntime.dll
-    COMMAND ${CMAKE_COMMAND} -E copy
-            ${GFX_PIX_PATH}/bin/x64/WinPixEventRuntime.dll
-            ${CMAKE_CURRENT_BINARY_DIR}/RelWithDebInfo/WinPixEventRuntime.dll)
+add_custom_command(TARGET 00-hellotriangle POST_BUILD
+    COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_RUNTIME_DLLS:00-hellotriangle> $<TARGET_FILE_DIR:00-hellotriangle>
+    COMMAND_EXPAND_LISTS
+)
```
maoliver-amd, line 22: Simplified the install commands as this is all that is needed

gfx.cpp
```diff
@@ -68,6 +68,17 @@ SOFTWARE.
...
+extern "C"
+{
+__declspec(dllexport) extern const UINT D3D12SDKVersion = 613;
+__declspec(dllexport) extern char8_t const *D3D12SDKPath = u8".\\";
+
+__declspec(dllexport) UINT GetD3D12SDKVersion()
+{
+    return D3D12SDKVersion;
+}
+}
```
maoliver-amd, line 80: The agility SDK variables are exported directly from gfx, this way any project using gfx doesnt have to be updated to define them if they didnt already. It does mean that any project that already was pulling in the agility sdk will have to remove their equivalent of these as otherwise they would clash (and it doesnt make sense for a parent project to pull in agility anymore as that could create versioning issues etc.).
The GetD3D12SDKVersion function is there to force the compiler to export these symbols as gfx is built as a static lib (and thus doesnt have exports) so we need to force exporting of these symbols from whatever exe/dll imports gfx

gboisse (approved):

PR #116, opened 2024-03-20
Increase max name length

VladimirMakeew:
Increasing `kGfxConstant_MaxNameLength` up to 260 characters (`MAX_PATH` from `windows.h` header file) should fix the problem with [long shader paths](https://github.com/Radeon-Pro/Capsaicin/issues/567).
The root of the problem is `GfxProgram::name` buffer that is too small to store long paths and the full path to the shader file is truncated (for example in `GfxInternal::createProgram`).

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -2309,7 +2309,7 @@ public:
         shader_model = (shader_model != nullptr ? shader_model : dxr_device_ != nullptr ? "6_5" : "6_0");
         Program &gfx_program = programs_.insert(program);
         gfx_program.shader_model_ = shader_model;
-        gfx_program.file_path_ = program.name;
+        gfx_program.file_path_ = name;
         gfx_program.cs_ = program_desc.cs;
         gfx_program.as_ = program_desc.as;
         gfx_program.ms_ = program_desc.ms;
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -2309,7 +2309,7 @@ public:
         shader_model = (shader_model != nullptr ? shader_model : dxr_device_ != nullptr ? "6_5" : "6_0");
         Program &gfx_program = programs_.insert(program);
         gfx_program.shader_model_ = shader_model;
-        gfx_program.file_path_ = name;
+        gfx_program.file_path_ = (name != nullptr) ? name : program.name;
         gfx_program.cs_ = program_desc.cs;
         gfx_program.as_ = program_desc.as;
         gfx_program.ms_ = program_desc.ms;
--- a/gfx_core.h
+++ b/gfx_core.h
@@ -52,7 +52,7 @@ enum GfxConstant
     kGfxConstant_BackBufferCount  = 3,
     kGfxConstant_MaxRenderTarget  = 8,
     kGfxConstant_MaxAnisotropy    = 8,
-    kGfxConstant_MaxNameLength    = 260,
+    kGfxConstant_MaxNameLength    = 64,
     kGfxConstant_NumBindlessSlots = 1024
 };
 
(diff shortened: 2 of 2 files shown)


gboisse:
Thanks Vladimir, however this "name" should not be used in any shader compilation-related process.
If it does, we should fix it.

The name of all `gfx` resources should only ever be used for debugging purposes (so it shows up in Pix for instance) and as a matter of fact can be renamed by the user of the API to anything.

Under the hood, the full name and path should be stored as strings under the internal `Program` structure:
https://github.com/gboisse/gfx/blob/7e89c1156aeb4dc8323cae5da343caa14fd980da/gfx.cpp#L645-L646

Let's make sure that these are used for all shader compilation functions and not that (re-namable) `GfxProgram::name`, which is, again, purely intended for debugging purposes.

gboisse:
Investigated this a bit further and I'd say this buffer is probably the culprit:
https://github.com/gboisse/gfx/blob/7e89c1156aeb4dc8323cae5da343caa14fd980da/gfx.cpp#L8116

Please turn both `shader_file` and `wshader_file` into a `std:string` and `std::wstring` respectively and we can call it a day.

VladimirMakeew:
Thanks for quick answer and investigation.
I was thinking about this code: https://github.com/gboisse/gfx/blob/master/gfx.cpp#L2312
Instead of creating `gfx_program.file_path` from argument we use `program.name`. program.name as you mentioned is a small buffer for debug purposes, so I guess I should fix this place first.

gboisse:
ha, good finding, you've definitely found another issue 😉

Yes, that's incorrect indeed and needs fixing.

gfx.cpp
```diff
@@ -2309,7 +2309,7 @@ class GfxInternal
         shader_model = (shader_model != nullptr ? shader_model : dxr_device_ != nullptr ? "6_5" : "6_0");
         Program &gfx_program = programs_.insert(program);
         gfx_program.shader_model_ = shader_model;
-        gfx_program.file_path_ = program.name;
+        gfx_program.file_path_ = name;
```
maoliver-amd, line 2312: Should be `gfx_program.file_path_ = (name != nullptr) ? name : program.name`
name may be nullptr in which case we can safely use program.name as in that case it is guaranteed to be valid and set to "gfx_Program%program.handle" which will be less than `kGfxConstant_MaxNameLength`

maoliver-amd:
> Investigated this a bit further and I'd say this buffer is probably the culprit:
> 
> https://github.com/gboisse/gfx/blob/7e89c1156aeb4dc8323cae5da343caa14fd980da/gfx.cpp#L8116
> 
> Please turn both `shader_file` and `wshader_file` into a `std:string` and `std::wstring` respectively and we can call it a day.
> 
> While you're at it, also apply a similar fix to the `.pdb` files please:
> 
> https://github.com/gboisse/gfx/blob/7e89c1156aeb4dc8323cae5da343caa14fd980da/gfx.cpp#L8296



> Investigated this a bit further and I'd say this buffer is probably the culprit:
> 
> https://github.com/gboisse/gfx/blob/7e89c1156aeb4dc8323cae5da343caa14fd980da/gfx.cpp#L8116
> 
> Please turn both `shader_file` and `wshader_file` into a `std:string` and `std::wstring` respectively and we can call it a day.
> 
> While you're at it, also apply a similar fix to the `.pdb` files please:
> 
> https://github.com/gboisse/gfx/blob/7e89c1156aeb4dc8323cae5da343caa14fd980da/gfx.cpp#L8296

This has already been done in #114

gfx.cpp
VladimirMakeew: Fixed.

maoliver-amd (approved):

PR #119, opened 2024-03-26
Support additional include paths for kernels

VladimirMakeew:
Allow users to pass additional shader include search paths when creating mesh, compute, graphics or raytracing kernels.
Parameters are optional, same as with defines.

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -736,6 +736,7 @@ class GfxInternal
         std::vector<String> defines_;
         std::vector<String> exports_;
         std::vector<String> subobjects_;
+        std::vector<String> include_paths_;
         std::map<std::wstring, LocalRootSignatureAssociation> local_root_signature_associations_;
         uint64_t descriptor_heap_id_ = 0;
         uint32_t *num_threads_ = nullptr;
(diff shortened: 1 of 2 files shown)


maoliver-amd:
Would it make sense to add the include paths in at the program level instead of kernel? Programs already have the shader path so include paths may make more sense there. This also reduces the number of places that include paths need to be added as its just in one location. This is more of a design decision so @gboisse what do you think?

gfx.cpp
```diff
@@ -2504,6 +2505,7 @@ class GfxInternal
         gfx_kernel.type_ = Kernel::kType_Mesh;
         gfx_kernel.draw_state_ = gfx_draw_state->draw_state_;
         for(uint32_t i = 0; i < define_count; ++i) gfx_kernel.defines_.push_back(defines[i]);
+        for (uint32_t i = 0; i < include_path_count; ++i) gfx_kernel.include_paths_.push_back(include_paths[i]);
```
maoliver-amd, line 2508: Style nit, shouldnt be a space between for and (

gfx.cpp
```diff
@@ -2623,6 +2628,7 @@ class GfxInternal
         for(uint32_t i = 0; i < define_count; ++i) gfx_kernel.defines_.push_back(defines[i]);
         for(uint32_t i = 0; i < export_count; ++i) gfx_kernel.exports_.push_back(exports[i]);
         for(uint32_t i = 0; i < subobject_count; ++i) gfx_kernel.subobjects_.push_back(subobjects[i]);
+        for (uint32_t i = 0; i < include_path_count; ++i) gfx_kernel.include_paths_.push_back(include_paths[i]);
```
maoliver-amd, line 2631: extra space again

VladimirMakeew:
Discussed it with @gboisse and we decided to pass additional shader include paths to programs instead of kernels.

PR #121, opened 2024-03-28
Add windowless mode support

picul:


gfx.cpp
```diff
@@ -96,13 +96,13 @@ class GfxInternal
     ID3D12GraphicsCommandList *command_list_ = nullptr;
     ID3D12GraphicsCommandList4 *dxr_command_list_ = nullptr;
     ID3D12GraphicsCommandList6 *mesh_command_list_ = nullptr;
-    ID3D12CommandAllocator **command_allocators_ = nullptr;
+    ID3D12CommandAllocator *command_allocators_[kGfxConstant_MaxBackBufferCount] = {};
```
gboisse, line 99: Why this change? These can't be statically-allocated arrays as another way of using `gfx` is by creating an "interop" context in which the user can supply an arbitrary value for `max_frames_in_flight`:
https://github.com/gboisse/gfx/blob/d617b252a68ea27116a5a26027530c22d45bc9e6/gfx.h#L450

In such a case, `gfx` doesn't own the number of backbuffers, etc. but has to align to the application being "interop'd" with and we can't rely on the `kGfxConstant_MaxBackBufferCount` constant.

gfx.cpp
```diff
@@ -1139,8 +1175,9 @@ class GfxInternal
         if(!SUCCEEDED(factory->EnumAdapterByLuid(device->GetAdapterLuid(), IID_PPV_ARGS(&adapter_))))
             return GFX_SET_ERROR(kGfxResult_InternalError, "Unable to find interop adapter");
 
+        is_interop_ = true;
         device_ = device;
-        max_frames_in_flight_ = GFX_MAX(max_frames_in_flight, 1u);
+        max_frames_in_flight_ = GFX_MIN(GFX_MAX(max_frames_in_flight, 1u), (uint32_t)kGfxConstant_MaxBackBufferCount);
```
gboisse, line 1180: This is incorrect, similarly to my previous message, we can't rely on that `kGfxConstant_MaxBackBufferCount` constant beyond being a default value during initialisation.

gboisse:
ha just realised that `kGfxConstant_MaxBackBufferCount` is something you added on top of the existing `kGfxConstant_BackBufferCount`, any reason for this?

Without a strong reason, I'd rather keep the previous version (with unbounded max. frames in flight) as it is less limiting when using the interop functionality.

gfx.h
```diff
@@ -48,7 +48,8 @@ enum GfxCreateContextFlag
 };
 typedef uint32_t GfxCreateContextFlags;
 
-GfxContext gfxCreateContext(HWND window, GfxCreateContextFlags flags = 0, IDXGIAdapter *adapter = nullptr);
+GfxContext gfxCreateContext(HWND window, GfxCreateContextFlags flags, IDXGIAdapter *adapter = nullptr);
+GfxContext gfxCreateContext(uint32_t window_width, uint32_t window_height, GfxCreateContextFlags flags, IDXGIAdapter *adapter = nullptr);
```
gboisse, line 52: Any reason for removing the default value on initialization flags? `GfxCreateContextFlags flags = 0`.
The lib's examples now fail to build: https://github.com/gboisse/gfx/tree/master/examples

gfx.cpp
```diff
@@ -118,7 +118,9 @@ class GfxInternal
     ID3D12CommandSignature *draw_mesh_signature_ = nullptr;
     std::vector<D3D12_RESOURCE_BARRIER> resource_barriers_;
     ID3D12Resource **back_buffers_ = nullptr;
+    D3D12MA::Allocation **back_buffer_allocations_ = nullptr;
     uint32_t *back_buffer_rtvs_ = nullptr;
+    bool is_interop_ = false;
```
gboisse, line 123: What's the reason behind needing this boolean vs. previous version?
Not a bad change at all but I'd like to better understand the extent of the changes.

gfx.cpp
```diff
@@ -2788,6 +2844,9 @@ class GfxInternal
     {
         if(isInterop())
             return GFX_SET_ERROR(kGfxResult_InvalidOperation, "Cannot clear backbuffer when using an interop context");
+        // No swap chain - no back buffer
+        if(!swap_chain_)
+            return kGfxResult_NoError;
```
gboisse, line 2849: Similarly to trying to clear the backbuffer from an "interop context", this should return an error to the user rather than failing/nop-ing silently.

gfx.cpp
```diff
@@ -4363,7 +4440,7 @@ class GfxInternal
         bound_depth_stencil_target_ = {};
     }
 
-    inline bool isInterop() const { return (swap_chain_ != nullptr ? false : true); }
+    inline bool isInterop() const { return is_interop_; }
```
gboisse, line 4443: ah I guess the boolean is added because otherwise a headless context would be treated as an interop one with this check, please confirm.

gboisse:
Changes are looking good, but please review the comments I've left on the PR, thanks.

gfx.cpp
gboisse: It seems that this boolean is never set to `true` anywhere actually.

gfx.cpp
picul: True, that's the reason boolean was added. And by checking if swapchain is null we can determine whether windowless (swapchain-less) mode is used.

gfx.h
picul: Well the default value was removed because otherwise compiler could not distinguish between overloads in case two arguments are provided. Unfortunately I didn't know examples existed; can we just modify them to accomodate to these changes?

picul: The way it's done in https://github.com/gboisse/gfx/pull/123 works fin though.

PR #125, opened 2024-04-10
CMake: Update to use cmakes inbuilt fetch instead of submodules

maoliver-amd:
This replaces submodules and other third party deps with cmakes inbuilt dependency fetching functions. 
This has a few advantages:
- Reduces the size of the repo as we no longer need to include and distribute any dependency binaries.
- Centralizes the dependency version information into a single location in the cmake file to make tracking and updating easier
- Prevents issues when swapping between branches with different submodules
- Allows pulling in deps from a package manager if available, skipping the need to pull down and compile. Tested and works with vcpkg (which is now included with visual studio as default)

Also fixed some issues with spaces in paths and updated deps to their latest versions


gfx_imgui.cpp
```diff
@@ -283,16 +287,7 @@ class GfxImGuiInternal
...
-        if(bd != nullptr && bd->hWnd != 0)
-        {
-            POINT pos = {};
-            GetCursorPos(&pos);
-            ImGui_ImplWin32_NewFrame();
-            ScreenToClient(bd->hWnd, &pos);
-            io.MousePos.x = (float)GFX_MIN(GFX_MAX(pos.x, (LONG)0), (LONG)gfxGetBackBufferWidth(gfx_)  - 1);
-            io.MousePos.y = (float)GFX_MIN(GFX_MAX(pos.y, (LONG)0), (LONG)gfxGetBackBufferHeight(gfx_) - 1);
-        }
+        ImGui_ImplWin32_NewFrame();
```
maoliver-amd, line 290: This code didnt appear to be needed, imgui appears to handle all this withing `ImGui_ImplWin32_NewFrame`

CMakeLists.txt
```diff
@@ -68,103 +51,222 @@ if(CMAKE_SYSTEM_PROCESSOR MATCHES "(x86)|(X86)|(amd64)|(AMD64)")
...
+    URL            "https://www.nuget.org/api/v2/package/WinPixEventRuntime/1.0.240308001"
+    SOURCE_DIR     "${GFX_PIX_PATH}"
+)
+FetchContent_MakeAvailable(WinPixEventRuntime)
+file(READ "${GFX_PIX_PATH}/include/WinPixEventRuntime/PIXEvents.h" PIXE_FILE_CONTENTS)
+string(REPLACE "#define PIX_CONTEXT_EMIT_CPU_EVENTS" "# define PIX_CONTEXT_EMIT_CPU_EVENTS\n# include \"AmdPix3.h\"\n# define PIX_AMD_EXT\n" PIXE_FILE_CONTENTS "${PIXE_FILE_CONTENTS}")
+string(REPLACE " PIXBeginEventOnContextCpu(destination" " RgpPIXBeginEventOnContextCpu(destination" PIXE_FILE_CONTENTS "${PIXE_FILE_CONTENTS}")
+string(REPLACE " PIXSetMarkerOnContextCpu(destination" " RgpPIXSetMarkerOnContextCpu(destination" PIXE_FILE_CONTENTS "${PIXE_FILE_CONTENTS}")
+string(REPLACE "destination = PIXEndEventOnContextCpu(" "RgpPIXEndEventOnContextCpu(destination, " PIXE_FILE_CONTENTS "${PIXE_FILE_CONTENTS}")
+file(WRITE "${GFX_PIX_PATH}/include/WinPixEventRuntime/PIXEvents.h" "${PIXE_FILE_CONTENTS}")
```
maoliver-amd, line 147: Auto patches WinPix with AMDPix3 on download, theres now no need to do it manually every time pix is updated

.gitmodules
```diff
@@ -1,3 +0,0 @@
-[submodule "third_party/D3D12MemoryAllocator"]
-	path = third_party/D3D12MemoryAllocator
-	url = https://github.com/GPUOpen-LibrariesAndSDKs/D3D12MemoryAllocator.git
```
gboisse, line 3: I suppose we can get rid of this file completely now.

CMakeLists.txt
gboisse: That's neat.
Of course, this'd break if the patching process ever changes in a future version of Pix?
Not a problem tho, we can update that snippet then.

gfx_imgui.cpp
```diff
@@ -24,11 +24,15 @@ SOFTWARE.
 
 #include "gfx_imgui.h"
 
+#ifdef GFX_IMGUI_SOURCE
```
gboisse, line 27: Why is this new condition needed?

gboisse (approved):
Approved but please reply to the few comments I left, thanks.

CMakeLists.txt
maoliver-amd: Yep, would need to be changed if the function names changed. We have the winpix version in the cmake file so whenever bumping that we would have to change the patching process if needed. This should be pretty rare though

.gitmodules
maoliver-amd: True, Ive now removed it from the repo

gfx_imgui.cpp
maoliver-amd: Difference between pulling it in via package management vs directly form the repo. Package managers will provide a lib whereas the source repo doesnt define a lib target and so gets pulled in via source

PR #129, opened 2024-05-07
Add error message when shader file not found.

maoliver-amd:
Currently we get a rather confusing error later on in the shader linking stage as we donet have an upfront error when the passed shader file cant be found.
So I added one.


gboisse:
Indeed, although the reasoning behind this is that in a typical pipeline, not all shader stages are required (for instance a graphics kernel that has only VS and no PS, or a mesh kernel that has no AS, etc.).

Wouldn’t this change now spam a lot of (invalid) errors when these shader files aren’t found?

maoliver-amd:
Ah yes your right, it apeear I jumped the gun there. Ill close this but we should probably have some error message when no shaders are found somewhere as the current error is rather non-obvious

PR #130, opened 2024-05-09
Add support for intersection shaders for HW study

SShikhali:
- Added a new primitive type: `Procedural`
- ~~Added a way to specify custom `Max Payload Size` and `Max Attribute Size` for RT pipeline.~~
- Added a function that returns true if a primitive type is `Procedural`.

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -2282,7 +2282,7 @@ public:
         return buildRaytracingPrimitive(raytracing_primitive, gfx_raytracing_primitive, false);
     }
 
-    GfxResult buildRaytracingProceduralPrimitive(GfxRaytracingPrimitive const &raytracing_primitive, GfxBuffer const &aabb_buffer, uint32_t aabb_stride, GfxBuildRaytracingPrimitiveFlags build_flags)
+    GfxResult buildRaytracingPrimitiveProcedural(GfxRaytracingPrimitive const &raytracing_primitive, GfxBuffer const &aabb_buffer, uint32_t aabb_stride, GfxBuildRaytracingPrimitiveFlags build_flags)
     {
         if(dxr_device_ == nullptr)
             return kGfxResult_InvalidOperation; // avoid spamming console output
@@ -2418,7 +2418,7 @@ public:
         return buildRaytracingPrimitive(raytracing_primitive, gfx_raytracing_primitive, true);
     }
 
-    GfxResult updateRaytracingProceduralPrimitive(GfxRaytracingPrimitive const &raytracing_primitive, GfxBuffer const &aabb_buffer, uint32_t aabb_stride)
+    GfxResult updateRaytracingPrimitiveProcedural(GfxRaytracingPrimitive const &raytracing_primitive, GfxBuffer const &aabb_buffer, uint32_t aabb_stride)
     {
         if(dxr_device_ == nullptr)
             return kGfxResult_InvalidOperation; // avoid spamming console output
@@ -9362,11 +9362,11 @@ GfxResult gfxRaytracingPrimitiveBuild(GfxContext context, GfxRaytracingPrimitive
     return gfx->buildRaytracingPrimitive(raytracing_primitive, index_buffer, vertex_buffer, vertex_stride, build_flags);
 }
 
-GfxResult gfxRaytracingProceduralPrimitiveBuild(GfxContext context, GfxRaytracingPrimitive raytracing_primitive, GfxBuffer aabb_buffer, uint32_t aabb_stride, GfxBuildRaytracingPrimitiveFlags build_flags)
+GfxResult gfxRaytracingPrimitiveBuildProcedural(GfxContext context, GfxRaytracingPrimitive raytracing_primitive, GfxBuffer aabb_buffer, uint32_t aabb_stride, GfxBuildRaytracingPrimitiveFlags build_flags)
 {
     GfxInternal *gfx = GfxInternal::GetGfx(context);
     if(!gfx) return kGfxResult_InvalidParameter;
-    return gfx->buildRaytracingProceduralPrimitive(raytracing_primitive, aabb_buffer, aabb_stride, build_flags);
+    return gfx->buildRaytracingPrimitiveProcedural(raytracing_primitive, aabb_buffer, aabb_stride, build_flags);
 }
 
 GfxResult gfxRaytracingPrimitiveSetTransform(GfxContext context, GfxRaytracingPrimitive raytracing_primitive, float const *row_major_4x4_transform)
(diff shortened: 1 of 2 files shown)


gfx.cpp
```diff
@@ -2163,6 +2206,35 @@ class GfxInternal
...
+        if(!acceleration_structure_handles_.has_handle(acceleration_structure.handle))
+        {
+            GFX_PRINT_ERROR(kGfxResult_InvalidParameter, "Cannot create a raytracing primitive using an invalid acceleration structure object");
+            return raytracing_primitive;
+        }
+        raytracing_primitive.handle = raytracing_primitive_handles_.allocate_handle();
+        RaytracingPrimitive &gfx_raytracing_primitive = raytracing_primitives_.insert(raytracing_primitive);
+        AccelerationStructure &gfx_acceleration_structure = acceleration_structures_[acceleration_structure];
+        gfx_raytracing_primitive.index_ = (uint32_t)gfx_acceleration_structure.raytracing_primitives_.size();
+        for (uint32_t i = 0; i < ARRAYSIZE(gfx_raytracing_primitive.transform_); ++i)
```
maoliver-amd, line 2228: style nit: rmeove space after `for`

gfx.cpp
```diff
@@ -471,6 +472,16 @@ class GfxInternal
...
+        struct
+        {
+            uint32_t build_flags_ = 0;
+            GfxBuffer bvh_buffer_ = {};
+            uint64_t bvh_data_size_ = 0;
+            uint32_t aabb_stride_ = 0;
+            GfxBuffer aabb_buffer_ = {};
+            GfxAccelerationStructure acceleration_structure_ = {};
+        }
+        aabb_;
```
maoliver-amd, line 484: I think the name of this should be `procedural_` instead of `aabb_`. The other structs above are all named after the corresponding type in the enum {Triangles, Instance, Procedural}. Also its more than just a standard AABB so best to avoid confusion with the name.

gfx.cpp
```diff
@@ -6683,6 +6813,88 @@ class GfxInternal
         return kGfxResult_NoError;
     }
 
+    GfxResult buildRaytracingPrimitive_AABB(GfxRaytracingPrimitive const &raytracing_primitive, RaytracingPrimitive &gfx_raytracing_primitive, bool update)
```
maoliver-amd, line 6816: naming should have AABB-> Procedural

gfx.cpp
```diff
@@ -779,6 +790,10 @@ class GfxInternal
         Parameter *parameters_ = nullptr;
         uint32_t parameter_count_ = 0;
         uint32_t vertex_stride_ = 0;
+
+        // These two params are used for RT pipelines. If one of them is set, the other one must be set as well
+        uint32_t max_payload_size_ = 0;
+        uint32_t max_attribute_size_ = 0;
```
SShikhali, line 796: It will be removed, since it can be configured from shaders

gboisse: Can we remove these yet?

gfx.cpp
```diff
@@ -2162,7 +2230,26 @@ class GfxInternal
         RaytracingPrimitive const &gfx_raytracing_primitive = raytracing_primitives_[raytracing_primitive];
         GfxAccelerationStructure const &acceleration_structure = getRaytracingPrimitiveAccelerationStructure(gfx_raytracing_primitive);
         if(acceleration_structure_handles_.has_handle(acceleration_structure.handle))
+        {
             acceleration_structures_[acceleration_structure].needs_rebuild_ = true;
+            if(gfx_raytracing_primitive.type_ == RaytracingPrimitive::kType_Procedural)
+            {
+                auto &primitives = acceleration_structures_[acceleration_structure].raytracing_primitives_;
+                if(gfx_raytracing_primitive.index_ < primitives.size() && primitives[gfx_raytracing_primitive.index_].handle == raytracing_primitive.handle)
```
SShikhali, line 2238: I noticed that a primitive is not being removed from the list of all primitives in AS, so this fixes the issue.
Previously, it wasn't a problem since the object were loaded and deleted at once during scene init. But with introduction of AABBs, that's not always that case.

I'm not sure if it's 100% correct to shift indices to the left so I put an `if` so that this code doesn't trigger in usual use cases to avoid other potential issues. What do you think about it? @gboisse @maoliver-amd 
```
--primitive_ahead.index_;
--primitive_ahead.instance_id_;
```

gfx.cpp
```diff
@@ -5160,9 +5290,10 @@ class GfxInternal
     void collect(AccelerationStructure const &acceleration_structure)
     {
         destroyBuffer(acceleration_structure.bvh_buffer_);
-        for(size_t i = 0; i < acceleration_structure.raytracing_primitives_.size(); ++i)
-            if(raytracing_primitive_handles_.has_handle(acceleration_structure.raytracing_primitives_[i].handle))
-                destroyRaytracingPrimitive(acceleration_structure.raytracing_primitives_[i]);
+        auto copy = acceleration_structure.raytracing_primitives_;
```
SShikhali, line 5293: Now a copy is required since `destroyRaytracingPrimitive` call might remove elements

gfx.cpp
SShikhali: Done

gfx.cpp
SShikhali: Done

gboisse:
This PR is looking real good, I just want to better understand why that "shifting left" is needed, did it actually cause a bug?

Otherwise, I noticed we're missing the handling of the new `RaytracingPrimitive::kType_Procedural` type in this switch statement:
https://github.com/gboisse/gfx/blob/b67b718697af031b40a67dd704cff9e7e05196f9/gfx.cpp#L2135-L2146

Which yields a bit the question of can you instantiate a procedural primitive?

I believe it should be supported as it's just about repeating the BLAS with another transform, but unsure if the D3D12 specs have anything to say about this...

gfx.cpp
```diff
@@ -2303,6 +2396,31 @@ class GfxInternal
...
+        if(gfx_raytracing_primitive.type_ != RaytracingPrimitive::kType_Triangles)
+            return GFX_SET_ERROR(kGfxResult_InvalidOperation, "Cannot update a non-triangle raytracing primitive object");
+        destroyBuffer(gfx_raytracing_primitive.triangles_.index_buffer_);
+        destroyBuffer(gfx_raytracing_primitive.triangles_.vertex_buffer_);
+        gfx_raytracing_primitive.triangles_.index_buffer_ = {};
+        gfx_raytracing_primitive.triangles_.index_stride_ = 0;
+        gfx_raytracing_primitive.triangles_.vertex_buffer_ = createBufferRange(vertex_buffer, 0, vertex_buffer.size);
+        gfx_raytracing_primitive.triangles_.vertex_stride_ = vertex_stride;
+        return buildRaytracingPrimitive(raytracing_primitive, gfx_raytracing_primitive, true);
+    }
```
gboisse, line 2422: Not really related to this PR, but I realised we were missing the API function for updating a raytracing primitive that uses a vertex buffer but no index buffer, so thought I'd just add it here.

gboisse:
So, investigating a bit more that "shift left" thing; this is the function that makes sure to "resolve" the array of raytracing primitives if any deletion has happened:
https://github.com/gboisse/gfx/blob/fb1b2c707882bf1a4627949e0786ed76389298fd/gfx.cpp#L6675-L6693

So, we shouldn't need that shift left logic in `GfxInternal::destroyRaytracingPrimitive()`.

Possibly, what would be a more proper fix is to simply remove these 2 API functions as they needlessly expose internal state from the library that the user really shouldn't rely on:
https://github.com/gboisse/gfx/blob/fb1b2c707882bf1a4627949e0786ed76389298fd/gfx.h#L190-L191

What do you think about this? would it be ok to remove these 2 functions?

gboisse:
I think also having the `GfxAccelerationStructure` automatically delete all `GfxRaytracingPrimitive` upon destruction is a design mistake too:
https://github.com/gboisse/gfx/blob/fb1b2c707882bf1a4627949e0786ed76389298fd/gfx.h#L186

Probably, we'd be better off removing that logic and let the user explicitly handle the lifetime of raytracing primitives.

PR #133, opened 2024-05-17
Custom payload size

SShikhali:


SShikhali:
No need as it can be configured from shaders

PR #134, opened 2024-05-20
Shader model 6_8 + experimental shaders support.

amd-pmacieje:
1. HLSL shader model 6_8 support.
2. Experimental shaders support.


maoliver-amd:
Looks ok except that we cant upgrade the dxc version at this time due to breaking compilation crashes. A fix is in the works but until then the 2024_03_29 version of dxc doesnt work with our code

PR #135, opened 2024-05-21
CMake: Improve detection of windows kits needed to find d3d12.

maoliver-amd:
VSCode wasnt working correctly as it failed to find d3d12.lib. This is in part due to cmake not natively knowing where to search for installed windows SDKs. Previously this worked for regular Visual Studio as it internally knows where the SDK is installed and so just passing "d3d12.lib" was enough as it would search for the lib within its know import locations. Outside of msbuild though cmake would fail to find d3d12 and would fail when run from command line or through another environment (anything not targetting an msbuild  target).

This PR updates the logic to explicitly search for and find the latest installed Windows Kit and use that to find d3d12 within cmake itself.


CMakeLists.txt
```diff
@@ -59,7 +59,26 @@ else()
 endif()
 
 target_include_directories(gfx PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
-target_link_libraries(gfx PRIVATE d3d12 dxgi)
+# Find the latest windows SDK needed for d3d12 libs
+if(NOT CMAKE_WINDOWS_KITS_10_DIR)
+    set(CMAKE_WINDOWS_KITS_10_DIR "C:/Program Files (x86)/Windows Kits/10")
+endif()
+if(CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION)
```
maoliver-amd, line 66: When building for a visual studio target, cmake defines this to the targetted kit version so we reuse to match and thus has no change in behaviour for VS builds.

CMakeLists.txt
```diff
@@ -59,7 +59,26 @@ else()
...
+    STRING(REGEX MATCH "[0-9]+\\.[0-9]+\\.[0-9]+\\.[0-9]+" WINDOWS_KIT_VERSION ${CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION})
+else()
+    set(WINDOWS_KIT_VERSION "10.0.0.0")
+    file(GLOB _KITS RELATIVE ${CMAKE_WINDOWS_KITS_10_DIR}/Lib ${CMAKE_WINDOWS_KITS_10_DIR}/Lib/*)
+    foreach(_KIT ${_KITS})
+        if(${_KIT} VERSION_GREATER_EQUAL ${WINDOWS_KIT_VERSION})
+            set(WINDOWS_KIT_VERSION ${_KIT})
+        endif()
+    endforeach()
```
maoliver-amd, line 75: When not building for VS we dont have a target version so instead search for the highest version available

gboisse (approved):

maoliver-amd:
Still needs #114 to fully fix stock compilation issues in VSCode

PR #136, opened 2024-05-30
Fix CodeQL errors/warnings.

maoliver-amd:
Mostly trivial fixes such as passing structs by reference instead of by value

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -1974,7 +1974,7 @@ public:
                 continue;   // invalid raytracing primitive object
             active_raytracing_primitives_.push_back(raytracing_primitive);
             RaytracingPrimitive const &gfx_raytracing_primitive = raytracing_primitives_[raytracing_primitive];
-            GfxBuffer buffer = getRaytracingPrimitiveBuffer(gfx_raytracing_primitive);
+            GfxBuffer const &buffer = getRaytracingPrimitiveBuffer(gfx_raytracing_primitive);
             if(!buffer_handles_.has_handle(buffer.handle))
                 continue;   // no valid BVH memory, probably wasn't built
             D3D12_RAYTRACING_INSTANCE_DESC instance_desc = {};
@@ -7410,7 +7410,7 @@ private:
                     }
                     else
                     {
-                        GfxBuffer buffer = parameter.parameter_->data_.buffer_.buffers_[j];
+                        GfxBuffer const &buffer = parameter.parameter_->data_.buffer_.buffers_[j];
                         if(!buffer_handles_.has_handle(buffer.handle))
                         {
                             if(buffer.handle != 0)
@@ -7473,7 +7473,7 @@ private:
                     }
                     else
                     {
-                        GfxBuffer buffer = parameter.parameter_->data_.buffer_.buffers_[j];
+                        GfxBuffer const &buffer = parameter.parameter_->data_.buffer_.buffers_[j];
                         if(!buffer_handles_.has_handle(buffer.handle))
                         {
                             if(buffer.handle != 0)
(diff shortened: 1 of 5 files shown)


gfx.cpp
```diff
@@ -1561,20 +1561,16 @@ class GfxInternal
...
-                gfx_buffer.resource_->Map(0, nullptr, &gfx_buffer.data_);
-                break;
+                D3D12_RANGE read_range = {};
+                gfx_buffer.resource_->Map(0, &read_range, &gfx_buffer.data_);
+                memset(gfx_buffer.data_, 0, (size_t)resource_desc.Width);
+                if(data)
+                    memcpy(gfx_buffer.data_, data, (size_t)size);
             }
+            else
+                gfx_buffer.resource_->Map(0, nullptr, &gfx_buffer.data_);
```
maoliver-amd, line 1573: switch statements with only 1 case are apparently frowned upon. So converted to simple if/else

gfx.cpp
```diff
@@ -2894,7 +2884,9 @@ class GfxInternal
         default:    // kGfxCpuAccess_None
             {
                 GfxKernel const bound_kernel = bound_kernel_;
-                setProgramBuffer(clear_buffer_program_, "OutputBuffer", buffer);
+                GfxBuffer       buffer_temp  = buffer;
+                buffer_temp.setStride(sizeof(uint32_t));
+                setProgramBuffer(clear_buffer_program_, "OutputBuffer", buffer_temp);
```
maoliver-amd, line 2889: Input buffer is now by ref so needed to create a copy that can be modified by changing stride (same functionaly as what was happening previously with pass by value)

gfx_imgui.h
```diff
@@ -37,7 +37,7 @@ SOFTWARE.
 //! ImGui initialization/termination.
 //!
 
-GfxResult gfxImGuiInitialize(GfxContext gfx, char const **font_filenames = nullptr, uint32_t font_count = 0,
+GfxResult gfxImGuiInitialize(GfxContext const &gfx, char const **font_filenames = nullptr, uint32_t font_count = 0,
```
gboisse, line 40: I think I'd rather keep these public API functions without the `const &` thing.
Not massively important I suppose, but equally that copy of `GfxContext` isn't realistically likely to cause any performance issue 🙂

maoliver-amd: The warnings are all about the size of the structs being larger than some threshold.
For reference:
sizeof(GfxContext) 80B
sizeof(GfxBuffer) 96B
sizeof(GfxTexture) 120B
sizeof(GfxAccelerationStructure) 80B

which is not huge but not trivially small either. The big size influence is all of these have a static char array as a member variable (fixed size at 65 chars) that has to be copied (which as its fixed size is potentially more costly to copy than strings that contain fewer chars). So you can see why analyzers would suggest they not be copied by value.

gboisse:
This PR has gotten quite out of sync. with `master` by now, should we merge with latest or simply close it?
I'm personally ok with closing it but please let me know if you have some another opinion.

maoliver-amd:
There were a couple of actual bug fixes so atleast we should cherry pick those out. The pass by reference when benchmarked provided a minor performance increase, mainly for debug builds where the pass by value copy wasnt optimised away and the program wasnt already GPU limited. Technically pass by constant reference can accept more input types than by value (for instance it can accept r-values whereas pass by value will compiler error) so it s an API change but one that shouldnt be noticed by any users. But as to whether we add that is up to you. I can update this PR with whatever changes are agreed upon

maoliver-amd:
Closing this as the bugs fixes have been applied in other PRs

PR #137, opened 2024-05-31
Improve GLTF loading support for compressed textures.

maoliver-amd:
This adds an additional material flag for exposing GLTF alpha mode values. These are values used to define the alpha blending mode for a material (i.e. opaque, blend or clip) and are intended to be used to determine if alpha testing should be performed on a surface and not just by checking if albedo has an alpha channel (which is done currently). Since for compressed textures (especially BC7) there is no method to actually detect if the texture contains non opaque alpha channel data then the alpha mode flag should be used instead.

This also fixes an issue with loading GLTF scenes that only contain compressed textures (either basisu ktx or dds) and not any uncompressed textures. In this case the previous behaviour would fail to match images to materials and textures would be displayed.

In addition to the above this also adds support for the AMD_split_roughnessMetallic GLTF extension that enables explicitly setting a separate texture for metallicity and roughness

### DIFF
--- a/gfx_scene.cpp
+++ b/gfx_scene.cpp
@@ -1364,6 +1364,13 @@ private:
                 material.clearcoat_roughness = gltf_material.clearcoat.clearcoat_roughness_factor;
             }
             if(gltf_material.double_sided) material.flags |= kGfxMaterialFlag_DoubleSided;
+            switch(gltf_material.alpha_mode)
+            {
+            case cgltf_alpha_mode_opaque: material.alpha_mode = GfxMaterialAlphaMode_Opaque; break;
+            case cgltf_alpha_mode_mask: material.alpha_mode = GfxMaterialAlphaMode_Mask; break;
+            case cgltf_alpha_mode_blend: material.alpha_mode = GfxMaterialAlphaMode_Blend; break;
+            default: material.alpha_mode = GfxMaterialAlphaMode_Opaque; break;
+            }
             cgltf_texture const *albedo_map_text = gltf_material_pbr.base_color_texture.texture;
             it = (albedo_map_text != nullptr ? images.find(albedo_map_text->image) : images.end());
             if(it != images.end())
--- a/gfx_scene.cpp
+++ b/gfx_scene.cpp
@@ -1268,34 +1268,49 @@ private:
             lights[&gltf_light] = light_ref;
         }
         std::map<cgltf_image const *, GfxConstRef<GfxImage>> images;
-        for(size_t i = 0; i < gltf_model->textures_count; ++i)
-        {
-            cgltf_texture const &gltf_texture = gltf_model->textures[i];
-            uint32_t texture_dds_index = std::numeric_limits<uint32_t>::max();
-            for(uint64_t e = 0; e < gltf_texture.extensions_count; ++e)
+        std::map<cgltf_texture const *, GfxConstRef<GfxImage>> textures;
+        auto get_best_image = [&](cgltf_texture const *texture) -> cgltf_image * {
+            if(texture == nullptr) return nullptr;
+            for(uint64_t i = 0; i < texture->extensions_count; ++i)
             {
-                if(strcmp(gltf_texture.extensions[e].name, "MSFT_texture_dds") == 0)
+                if(strcmp(texture->extensions[i].name, "MSFT_texture_dds") == 0)
                 {
-                    std::string json = gltf_texture.extensions[e].data;
-                    json.erase(remove_if(json.begin(), json.end(),
-                                   [](char c) { return (c == '\r' || c == '\t' || c == ' ' || c == '\n'); }),
+                    // Need to get the dds out of the extension
+                    std::string json = texture->extensions[i].data;
+                    json.erase(
+                        remove_if(json.begin(), json.end(),
+                            [](char c) { return (c == '\r' || c == '\t' || c == ' ' || c == '\n'); }),
                         json.end());
                     constexpr std::string_view source_tag = "\"source\":";
                     if(auto pos = json.find(source_tag); pos != std::string::npos)
                     {
                         pos += source_tag.length();
-                        std::string clipped = json.substr(pos, json.find_first_not_of("0123456789", pos + 1) - pos);
-                        texture_dds_index = stoi(clipped);
+                        std::string clipped =
+                            json.substr(pos, json.find_first_not_of("0123456789", pos + 1) - pos);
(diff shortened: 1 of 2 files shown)


gfx_scene.cpp
```diff
@@ -1268,34 +1268,49 @@ class GfxSceneInternal
...
-            cgltf_texture const &gltf_texture = gltf_model->textures[i];
-            uint32_t texture_dds_index = std::numeric_limits<uint32_t>::max();
-            for(uint64_t e = 0; e < gltf_texture.extensions_count; ++e)
-            {
-                if(strcmp(gltf_texture.extensions[e].name, "MSFT_texture_dds") == 0)
-                {
-                    std::string json = gltf_texture.extensions[e].data;
-                    json.erase(remove_if(json.begin(), json.end(),
-                                   [](char c) { return (c == '\r' || c == '\t' || c == ' ' || c == '\n'); }),
+        std::map<cgltf_texture const *, GfxConstRef<GfxImage>> textures;
```
maoliver-amd, line 1271: Added an additional texture lookup for matching gltf textures with loaded gfx images. This is used as multiple gltf textures may reference the same gltf image. looking up the correct gltf image requires checking for presence of basisu or dds textures so for simplicity and speed we cache the texture to gfx image conversion.

gfx_scene.cpp
```diff
@@ -1364,131 +1381,182 @@ class GfxSceneInternal
                 material.clearcoat_roughness = gltf_material.clearcoat.clearcoat_roughness_factor;
             }
             if(gltf_material.double_sided) material.flags |= kGfxMaterialFlag_DoubleSided;
+            switch(gltf_material.alpha_mode)
+            {
+            case cgltf_alpha_mode_opaque: material.alpha_mode = GfxMaterialAlphaMode_Opaque; break;
+            case cgltf_alpha_mode_mask: material.alpha_mode = GfxMaterialAlphaMode_Mask; break;
+            case cgltf_alpha_mode_blend: material.alpha_mode = GfxMaterialAlphaMode_Blend; break;
+            default: material.alpha_mode = GfxMaterialAlphaMode_Opaque; break;
+            }
```
maoliver-amd, line 1390: The new alpha mode loading

gfx_scene.cpp
```diff
@@ -1364,131 +1381,182 @@ class GfxSceneInternal
...
             cgltf_texture const *albedo_map_text = gltf_material_pbr.base_color_texture.texture;
-            it = (albedo_map_text != nullptr ? images.find(albedo_map_text->image) : images.end());
-            if(it != images.end())
+            it = (albedo_map_text != nullptr ? textures.find(albedo_map_text) : textures.end());
+            if(it != textures.end())
```
maoliver-amd, line 1393: The texture map is now used for image lookups as searching by image will fail when the base `->image` member is null. This occurs when a scene file only contains basisu or dds textures and not any uncompressed variants

gfx_scene.cpp
```diff
@@ -1364,131 +1381,182 @@ class GfxSceneInternal
...
+                        has_split_extension = true;
+                    }
                 }
-                else
+            }
+            if (!has_split_extension)
+            {
+                cgltf_texture const *metallicity_roughness_map_text = gltf_material_pbr.metallic_roughness_texture.texture;
+                it = (metallicity_roughness_map_text != nullptr ? textures.find(metallicity_roughness_map_text) : textures.end());
+                if(it != textures.end())
```
maoliver-amd, line 1445: The diff below this point is juts an indent of the existing code to be within the if check. We only search for regular metallicRoughness texture(s) should the extension not be found.

gfx_scene.cpp
```diff
@@ -2630,7 +2698,11 @@ class GfxSceneInternal
                 mipDepth = GFX_MAX(1U, mipDepth / 2);
             }
         }
-        image_ref->flags = (image_ref->channel_count != 4 ? 0 : kGfxImageFlag_HasAlphaChannel);
+        image_ref->flags = (image_ref->channel_count != 4
+            || (image_ref->format == DXGI_FORMAT_BC7_TYPELESS
+            || image_ref->format == DXGI_FORMAT_BC7_UNORM
+            || image_ref->format == DXGI_FORMAT_BC7_UNORM_SRGB) //BC7 may or may not have alpha
+            ? 0 : kGfxImageFlag_HasAlphaChannel);
```
maoliver-amd, line 2705: We now only set the has alpha channel flag if we are sure one exists. BC7 can either have one or not depending on the compression mode so there is no way of knowing. Instead of always flagging an alpha we default to not flagging it in the case of BC7, this may still be wrong but if using BC7 albedo one should use the new alpha_mode flag to indicate of alpha testing should be performed

gfx_scene.h
```diff
@@ -311,6 +311,15 @@ enum GfxMaterialFlag
...
 typedef uint32_t GfxMaterialFlags;
 
+enum GfxMaterialAlphaMode
+{
+    GfxMaterialAlphaMode_Opaque = 0,
+    GfxMaterialAlphaMode_Blend = 1,
+    GfxMaterialAlphaMode_Mask = 2,
+
+    GfxMaterialAlphaMode_Count
+};
```
maoliver-amd, line 321: GLTF exposes 3 alpha modes: opaque, blend and clipped. Clipped also has an optional clip value (i.e. any alpha below some pre-sepcified value is considered clip) but we are currently just assuming the GLTF specs default of 0.5.

gfx_scene.cpp
```diff
@@ -1364,131 +1381,182 @@ class GfxSceneInternal
...
+                                    }
+                                }
+                            }
+                        };
+                        constexpr std::string_view metalTag = "\"metallicTexture\":{\"index\":";
+                        constexpr std::string_view roughTag = "\"roughnessTexture\":{\"index\":";
+                        get_texture(metalTag, material_ref->roughness_map);
+                        get_texture(roughTag, material_ref->metallicity_map);
+                        has_split_extension = true;
+                    }
```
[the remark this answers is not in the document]
gboisse: Sounds good!
Arguably, we could also expose a `GfxMaterialFlag` along the lines of `kGfxMaterialFlag_IsReflectanceCombined` or something and handle this accordingly at the app level (then we don't diverge from the core glTF specs iiuc).
But looking good otherwise.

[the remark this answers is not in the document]
maoliver-amd: We already diverged in that we have always been using split metal and roughness textures. Previously though we looked for them based on the name of the stock gltf combined metal/rough and failed if we couldnt find them. This extension basically allows for these files to be explicitly set without having the old behaviour of guessing the filename

PR #141, opened 2024-08-13
Set AMD and NVIDIA GPUs to the same adaptor score.

maoliver-amd:
Currently we are prioritising AMD GPUS with a higher adapter score. This has issues with newer Zen 4 and 5 CPUs as they contain a iGPU which is being given a higher priority over discrete cards from other manufacturers. This is a simple change that gives AMD and NVIDIA the same score which then relies on the physical memory check to then order them correctly.
Fixes an issue on Zen5 CPU with NVIDIA GPU whereby iGPU was always being selected.

I also added support for using `EnumAdapterByGpuPreference` which will use windows inbuilt gpu sorting to order adapters by performance

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -1083,7 +1083,7 @@ public:
                     switch(adapter_desc.VendorId)
                     {
                     case 0x1002u:   // AMD
-                        adapter_score = 3;
+                        adapter_score = 2;
                         break;
                     case 0x10DEu:   // NVIDIA
                         adapter_score = 2;
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -33,6 +33,7 @@ SOFTWARE.
 #include <dxcapi.h>             // shader compiler
 #include <d3d12shader.h>        // shader reflection
 #include <D3D12MemAlloc.h>
+#include <dxgi1_6.h>            // IDXGIFactory6
 
 #ifdef __clang__
 #   pragma clang diagnostic push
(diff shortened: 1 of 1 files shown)


gboisse (approved):

gfx.cpp
```diff
@@ -1001,15 +1002,15 @@ class GfxInternal
 
     GfxResult initialize(uint32_t width, uint32_t height, GfxCreateContextFlags flags, IDXGIAdapter *adapter, GfxContext &context)
     {
-        IDXGIFactory4 *factory = nullptr;
+        IDXGIFactory1 *factory = nullptr;
         if(!SUCCEEDED(CreateDXGIFactory1(IID_PPV_ARGS(&factory))))
```
maoliver-amd, line 1006: Use IDXGIFactory1 directly instead of implicit cast. Not needed specifically but picks up errors on older windows platforms

gfx.cpp
```diff
@@ -1066,45 +1070,60 @@ class GfxInternal
...
+                            adapter_score = 2;
+                            break;
+                        default:
+                            adapter_score = 1;
+                            break;
+                        }
+                    for(j = 0; j < i; ++j)
+                        if(adapter_score > adapter_scores[j] ||
+                           adapter_desc.DedicatedVideoMemory > adapter_descs[j].DedicatedVideoMemory ||
+                           adapter_desc.SharedSystemMemory > adapter_descs[j].SharedSystemMemory)
```
maoliver-amd, line 1115: Also extended the memory check to also order by shared memory which is a fallback for laptop GPUs

gfx.cpp
gboisse: ha, good catch!

PR #142, opened 2024-08-14
Add High bit depth detection and support

maoliver-amd:
Adds code to detect the current bit depth support for the connected display and sets the swapchain format accordingly. Currently it just detects 8b or 10b SDR support and automatically uses 10b if supported. This happens seamlessly and doesnt require any user changes to their app as d3d will automatically handle conversion when presenting buffers (standard float [0-1) to uint conversion).
Also added some code lines (currently commented out) that show how to detect HDR capable displays which can be used for any future HDR support updates.

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -33,6 +33,7 @@ SOFTWARE.
 #include <dxcapi.h>             // shader compiler
 #include <d3d12shader.h>        // shader reflection
 #include <D3D12MemAlloc.h>
+#include <dxgi1_6.h>             // IDXGIOutput6
 
 #ifdef __clang__
 #   pragma clang diagnostic push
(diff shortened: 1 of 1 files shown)


gfx.cpp
```diff
@@ -965,11 +965,29 @@ class GfxInternal
...
+            output->QueryInterface(&output6);
+            DXGI_OUTPUT_DESC1 output_desc;
+            output6->GetDesc1(&output_desc);
+            if(output_desc.BitsPerColor > 8)
+                output_format = DXGI_FORMAT_R10G10B10A2_UNORM;
+            /*if(output_desc.ColorSpace == DXGI_COLOR_SPACE_RGB_FULL_G2084_NONE_P2020)
+            {
+                output_type = DXGI_COLOR_SPACE_RGB_FULL_G2084_NONE_P2020;
+                output_format = DXGI_FORMAT_R10G10B10A2_UNORM;
+            }*/
```
maoliver-amd, line 983: These are the relevant lines for enabling a HDR display output. This should be wrapped in a user option as we dont want to auto set HDR as it requires user side changes. I didnt add such as user option as I have no way to test it and didnt know how you prefered to do it. The user option would be optional (i.e. enable HDR if available) which would require an additional function/return that can be used to check if HDR was in use. This is complicated further by windows unfortunate handling of multiple displays as it possible for a window to be dragged between displays were only 1 is HDR requiring additional code to check for this and then change the swapchain accordingly and notify the user. This still falls down though when a window is straddling 2 monitors where only 1 is HDR. So additional code is required if HDR is to be supported in windowed mode or when windows are draggable

gfx.cpp
```diff
@@ -965,11 +966,33 @@ class GfxInternal
...
+                if(output_desc.BitsPerColor > 8)
+                    back_buffer_format_ = DXGI_FORMAT_R10G10B10A2_UNORM;
+                //if(output_desc.ColorSpace == DXGI_COLOR_SPACE_RGB_FULL_G2084_NONE_P2020)
+                //{
+                //    color_space = DXGI_COLOR_SPACE_RGB_FULL_G2084_NONE_P2020;
+                //    back_buffer_format_ = DXGI_FORMAT_R10G10B10A2_UNORM;
+                //}
+                output6->Release();
+            }
+            output->Release();
```
gboisse, line 988: Fixed missing calls to `Release()`.

gfx.cpp
```diff
@@ -121,6 +121,7 @@ class GfxInternal
     std::vector<D3D12_RESOURCE_BARRIER> resource_barriers_;
     ID3D12Resource **back_buffers_ = nullptr;
     D3D12MA::Allocation **back_buffer_allocations_ = nullptr;
+    DXGI_FORMAT back_buffer_format_ = DXGI_FORMAT_R8G8B8A8_UNORM;
```
gboisse, line 124: Made this into a member variable so we don't have to re-fetch that `DXGI_SWAP_CHAIN_DESC1` structure each time.

gboisse (approved):

gfx.cpp
maoliver-amd: yep, good catch

PR #143, opened 2024-08-14
Fix clearing large buffers

gboisse:
Maximum number of groups is limited to `65535` on AMD for a given dispatch.

This was making clearing large buffers fail, this PR fixes it.


maoliver-amd (approved):
yep, works nicely

PR #145, opened 2024-10-01
Enable assert macros on release

pawkubik:
The current `master` fails on Release mode with MSVC:

```
[build] gfx.cpp(8413): error C2220: the following warning is treated as an error
[build] gfx.cpp(7716) : warning C6011: Dereferencing NULL pointer 'parameter.parameter_'. : Lines: 7704, 7705, 7706, 7708, 7711, 7712, 7713, 7714, 7715, 7716
```

There is `GFX_ASSERT(parameter.parameter_ != nullptr)` macro that stops the flow of the function and avoids the dereferencing, but it was only enabled in the Debug mode.

I changed it to occur in both Debug and Release and changed `GFX_BREAKPOINT` to no-op in Release mode.

### DIFF
--- a/gfx_core.h
+++ b/gfx_core.h
@@ -60,10 +60,6 @@ enum GfxConstant
 //! Public macros.
 //!
 
-#define GFX_ASSERT(X)   // skip it
-
-#define GFX_ASSERTMSG(X, ...)
-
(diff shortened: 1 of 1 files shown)


gboisse:
Superseded by https://github.com/gboisse/gfx/pull/153.

PR #146, opened 2024-10-03
Add functions for getting current display data

maoliver-amd:
Adds functions for checking the current display color space and depth as well as monitor specific information. This data is required for more advanced tonemapping operators or anything else that takes the actual display capabilities into account


gfx.cpp
```diff
@@ -967,8 +968,28 @@ class GfxInternal
...
+                if(intersect_area > best_area)
+                {
+                    output    = current_output;
+                    best_area = intersect_area;
+                }
+            }
+            if(current_output != output) current_output->Release();
+            output_i++;
+        }
+        if(output != nullptr)
```
maoliver-amd, line 992: Improved display detection when user has multiple monitors and the window is spanning multiple displays

PR #147, opened 2024-10-03
Quick shader cache

gboisse:


gfx.cpp
```diff
@@ -8849,6 +8856,86 @@ class GfxInternal
...
+                std::map<uint64_t, Shader>::const_iterator const it = shaders_.find(shader_key);
+                if(it != shaders_.end())
+                {
+                    shader_bytecode = (*it).second.shader_bytecode_;
+                    reflection = (*it).second.shader_reflection_;
+                    GFX_ASSERT(shader_bytecode != nullptr && reflection != nullptr);
+                    if(dxc_source) dxc_source->Release();
+                    return; // done
+                }
+                std::string shader_key_file = "./shader_cache/";
```
maoliver-amd, line 8897: MiniNN already creates a cache folder called "cache", the file extensions and naming are different so it may be usefull to reuse the same directory to keep things neat

maoliver-amd (approved):
Apart from the folder name, I tested it and it seems to work nicely. A few minor changes to source were correctly detected and recompiled while loading cached shaders was much faster

gfx.cpp
gboisse: Yea, that's a good point, and I thought of this too.
On another hand, it's kinda neat that the 2 caches are separate (HIP vs. D3D12) as they store very distinct artefacts after all?
Not completely sure which I prefer on that one 😛

PR #148, opened 2024-10-07
Add functions to swap buffers and textures.

maoliver-amd:
Useful utility functions to allow swapping buffers and textures. This just swaps member variables related to the stored data but preserve things like the buffer/texture name. This is useful for avoiding copy commands when copying from a temp buffer to an output as now the handles can just be swapped

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -3126,6 +3126,15 @@ public:
         return kGfxResult_NoError;
     }
 
+    GfxResult encodeSwapBuffer(GfxBuffer &buf1, GfxBuffer &buf2)
+    {
+        std::swap(buf1.handle, buf2.handle);
+        std::swap(buf1.size, buf2.size);
+        std::swap(buf1.stride, buf2.stride);
+        std::swap(buf1.cpu_access, buf2.cpu_access);
+        return kGfxResult_NoError;
+    }
+
     GfxResult encodeClearBuffer(GfxBuffer buffer, uint32_t clear_value)
     {
         GfxResult result = kGfxResult_NoError;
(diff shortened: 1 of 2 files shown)


gboisse:
Any reason you couldn't simply do `std::swap(buf1, buf2)` directly in the app?
We can have this utility but I don't think I would put in under `gfxCommandXxx()` since this is meant to signify actions that will happen later on once the GPU processes the command stream when this swap actually happens straight away on the CPU, so not really a GPU command.

maoliver-amd:
`std::swap` will copy the texture/buffer names as well. The idea with this is that is only copies the underlying resource but the name stays the same. So if you have a render pass that reads from a texture named "Output" for instance and outputs to a temp texture "Temp" (as the pass reads multiple input pixels for each output so cant directly overwrite). The since "Output" is your actual output texture you then need to copy the contents of "Temp" into "Output". So this PR adds functions that allow you to avoid that by just swapping the resource, so now "Output" still has the correct name but contains the contents of "Temp" without having to a copy operation.

As for the name, thats easy to change. If you have any suggestions then Ill change it to match (other wise ill just remove "Command" and leave it at that)

maoliver-amd:
Ahh, ok thats problematic. Were is the code that does the name set on the resource as ill add that to the new function to ensure it is in sync as well. Im assuming that updating the resource name will still be faster than doing a complete texture copy. We have a couple of locations that ive found where we can sneak some performance by just doing a swap instead of a full copy

gboisse:
I'm a bit confused, we can still do the swap in the app. without needing any texture copy no?
Simply that the name changes but I'm not sure what this PR is solving w.r.t. copying or not GPU data?

maoliver-amd:
OK ive updated with the function name change and also cleared the resource name flag so the name is updated for the resource.

> I'm a bit confused, we can still do the swap in the app. without needing any texture copy no? Simply that the name changes but I'm not sure what this PR is solving w.r.t. copying or not GPU data?

Yep we can still do a swap and it will work just the names wont match, this PR just lets you swap while ensuring the names match. Currently we have some AOVs (e.g. "DepthBuffer", "PreviousDepthBuffer") that we do a GPU copy from the first to the second every frame, we can skip that copy by just doing a swap but previously that swap meant the names were out of sync which gets rather confusing to debug. This PR fixes that by allowing swapping while preserving names which means those copies can be removed

PR #149, opened 2024-10-07
Dont pass on key events to gfx when they are captured by ImGui.

maoliver-amd:
Currently key events are passed to both imgui and to gfx. This crates issues when commands are handled by both imgui and the application. An example is when typing into a imgui text field and the keyboard commands are not being properly caught and are still passed to the app (e.g. typing wasd into text field results in camera movement as well).

This PR corrects this by detecting when ImGui is capturing keyboard input and only passing it on if its not used by imgui

### DIFF
--- a/gfx_window.cpp
+++ b/gfx_window.cpp
@@ -267,23 +267,40 @@ private:
                 }
                 break;
             case WM_CHAR:
-            case WM_SETCURSOR:
-            case WM_DEVICECHANGE:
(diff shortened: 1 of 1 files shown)


maoliver-amd:
OK turns out this can break things when a key is being held down while focus changes to ImGui. A workaround could be to always send keyup events to gfx but instead ill just close this and rely and manually checking the imgui capture status in programs.

PR #151, opened 2024-10-08
Workaround for dispatches exceeding max. num groups on AMD

gboisse:
Implemented directly at dispatch command level; should fix all issues of this kind 🙂


maoliver-amd (approved):
looks good

PR #155, opened 2024-12-13
Add detection of device failures.

maoliver-amd:
This adds functionality to identify device failures such as driver timeouts, device removal or device change (such as in multi-gpu systems). It will print an error log and uses a new return value to indicate to the user that a failure has occurred.
This can be used by the user to either exit or reset the current gfx context and continue.

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -9623,7 +9623,7 @@ GfxResult gfxDestroyContext(GfxContext context)
     return kGfxResult_NoError;
 }
 
-bool gfxIsValid(GfxContext context)
+bool gfxContextIsValid(GfxContext context)
 {
     GfxInternal *gfx = GfxInternal::GetGfx(context);
     if(!gfx) return false;
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -1430,13 +1430,21 @@ public:
     GfxResult terminate()
     {
         destroyBuffer(draw_id_buffer_);
+        draw_id_buffer_ = {};
         destroyKernel(clear_buffer_kernel_);
+        clear_buffer_kernel_ = {};
         destroyProgram(clear_buffer_program_);
+        clear_buffer_program_ = {};
         destroyKernel(copy_to_backbuffer_kernel_);
+        copy_to_backbuffer_kernel_ = {};
         destroyProgram(copy_to_backbuffer_program_);
+        copy_to_backbuffer_program_ = {};
         destroyBuffer(texture_upload_buffer_);
+        texture_upload_buffer_ = {};
         destroyBuffer(raytracing_scratch_buffer_);
+        raytracing_scratch_buffer_ = {};
         destroyBuffer(sort_scratch_buffer_);
+        sort_scratch_buffer_ = {};
 
         if(constant_buffer_pool_ != nullptr)
             for(uint32_t i = 0; i < max_frames_in_flight_; ++i)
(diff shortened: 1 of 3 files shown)


gfx.cpp
```diff
@@ -1502,84 +1527,162 @@ class GfxInternal
...
         if(draw_mesh_signature_ != nullptr)
+        {
             draw_mesh_signature_->Release();
+            draw_mesh_signature_ = nullptr;
+        }
+
+#ifdef PIX_AMD_EXT
+        tls_pAmdExtDeviceObject = nullptr;
+        tls_checkAmdDriver = true;
+#endif
```
maoliver-amd, line 1676: When AMDPIX is in use it needs to be properly reset as otherwise it will fail on re-initialize. Its not documented but this is how its done

gfx.cpp
```diff
@@ -1502,84 +1527,162 @@ class GfxInternal
...
 
         return kGfxResult_NoError;
     }
 
+    bool isValid() const
+    {
+        return device_ != nullptr && mem_allocator_ != nullptr && device_->GetDeviceRemovedReason() == S_OK;
+    }
```
maoliver-amd, line 1684: Added a new function that allows to check if a gfx context has been correctly initialized and that it is still valid (i.e. device is still valid)

gfx.cpp
```diff
@@ -9414,13 +9525,40 @@ class GfxInternal
...
+    {
+        HRESULT const reason = device_->GetDeviceRemovedReason();
+        LPSTR error_text = nullptr;
+        DWORD const result = FormatMessageA(FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_ALLOCATE_BUFFER, nullptr, reason,
+                MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), reinterpret_cast<LPSTR>(&error_text), 0, nullptr);
+        if (command_list_ != nullptr)
+        {
+            command_list_->Release();
+            command_list_ = nullptr;
+        }
```
maoliver-amd, line 9552: Need to explicitly close the command list here as otherwise will get failures later on even if just terminating

maoliver-amd:
For reference this can be tested using the VS command prompt run as administrator and using the command `dxcap -forcetdr` to force a device timeout removal

gfx.h
```diff
@@ -53,6 +53,7 @@ typedef uint32_t GfxCreateContextFlags;
 GfxContext gfxCreateContext(HWND window, GfxCreateContextFlags flags = 0, IDXGIAdapter *adapter = nullptr);
 GfxContext gfxCreateContext(uint32_t width, uint32_t height, GfxCreateContextFlags flags, IDXGIAdapter *adapter = nullptr);
 GfxResult gfxDestroyContext(GfxContext context);
+bool      gfxIsValid(GfxContext context);
```
gboisse, line 56: What about renaming this to `gfxHasDeviceCrashes()` or something along these lines?
`isValid` seems a bit vague imo, what do you think?

maoliver-amd: Well it returns whether gfx has been initialized and whether the device is still working so its more than just a device crash check. So `gfxIsContextValid` or something else, ideas welcom

gboisse: Got it, then let's name it `gfxContextIsValid()` just to be more consistent with the rest of the naming; i.e., it's `gfxCreateBuffer()` and `gfxDestroyBuffer()` but any other buffer-related function is named `gfxBufferXxx()`.

maoliver-amd: ok done

gboisse (approved):

PR #156, opened 2025-01-09
Format debug output to match visual studio diagnostic messages.

maoliver-amd:
This changes the format of output messages to match the format required by Visual Studio diagnostic messaging ([https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-diagnostic-format-for-tasks?view=vs-2022](https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-diagnostic-format-for-tasks?view=vs-2022))

This allows Visual Studio to detect the format of the messages and allows for quick navigation to the source of the message by simply double clicking the message in the output window. It also allows it to be picked up by the error window whcih can also be used for quick navigation. This works for shader compilation messages allowing for quickly traversing to the exact line and file the compilation message was generated for.

This requires a newer DXC version to support so this PR is based off of dxc-sm6.8 branch

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -2617,7 +2617,7 @@ public:
         char const last_char = file_path[strlen(file_path) - 1];
         char const *path_separator = (last_char == '/' || last_char == '\\' ? "" : "/");
         GFX_SNPRINTF(program.name, sizeof(program.name), "%s%s%s", file_path, path_separator, file_name);
-        shader_model = (shader_model != nullptr ? shader_model : dxr_device_ != nullptr ? "6_6" : "6_0");
+        shader_model = (shader_model != nullptr ? shader_model : dxr_device_ != nullptr ? "6_5" : "6_0");
         program.handle = program_handles_.allocate_handle();
         Program &gfx_program = programs_.insert(program);
         gfx_program.shader_model_ = shader_model;
@@ -2635,7 +2635,7 @@ public:
             GFX_SNPRINTF(program.name, sizeof(program.name), "%s", name);
         else
             GFX_SNPRINTF(program.name, sizeof(program.name), "gfx_Program%llu", program.handle);
-        shader_model = (shader_model != nullptr ? shader_model : dxr_device_ != nullptr ? "6_6" : "6_0");
+        shader_model = (shader_model != nullptr ? shader_model : dxr_device_ != nullptr ? "6_5" : "6_0");
         Program &gfx_program = programs_.insert(program);
         gfx_program.shader_model_ = shader_model;
         gfx_program.file_path_ = (name != nullptr ? name : program.name);
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -9007,6 +9007,7 @@ private:
         dxc_utils_->CreateReflection(&reflection_data, IID_PPV_ARGS(&reflection));
 
         if(shader_key != 0 && dxc_bytecode != nullptr && reflection != nullptr)
+        {
             if constexpr(std::is_same<ID3D12ShaderReflection, REFLECTION_TYPE>::value)
             {
                 GFX_ASSERT(shaders_.find(shader_key) == shaders_.end());
@@ -9026,6 +9027,7 @@ private:
                     fclose(fd); // write out reflection for shader caching
                 }
             }
+        }
         if(reflection) shader_bytecode = dxc_bytecode;
         if(!reflection) dxc_bytecode->Release();
         if(dxc_pdb_name) dxc_pdb_name->Release();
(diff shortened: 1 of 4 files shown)


gfx.cpp
```diff
@@ -2621,7 +2622,8 @@ class GfxInternal
         Program &gfx_program = programs_.insert(program);
         gfx_program.shader_model_ = shader_model;
         gfx_program.file_name_ = file_name;
-        gfx_program.file_path_ = file_path;
+        std::string const absolute_path = absolute(std::filesystem::path(file_path)).string();
+        gfx_program.file_path_ = absolute_path.c_str();
```
maoliver-amd, line 2626: Shader paths need to be absolute in order for the path detection to correctly work

gfx_core.h
```diff
@@ -198,58 +198,52 @@ static inline char const *gfxResultGetString(GfxResult result)
...
+    int32_t size = snprintf(nullptr, 0, "%s(%-4u): %s", file_name, line_number, format);
+    std::vector<char> body(size + 1);
+    snprintf(body.data(), body.size(), "%s(%-4u): %s", file_name, line_number, format);
+    size = vsnprintf(nullptr, 0, body.data(), args);
+    std::vector<char> message(size + 2);
+    vsnprintf(message.data(), message.size(), body.data(), args);
     va_end(args);
+    message[message.size() - 2] = '\n';
+    OutputDebugStringA(message.data());
+    puts(message.data());
```
maoliver-amd, line 212: I also updated the `PRINTLN` functions so that the output string is dynamically allocated. Previous implementation often cut of the filename due to `header` being to small and most dxc compiler output was cutoff due to body and message being to small

gfx.cpp
```diff
@@ -8882,6 +8884,11 @@ class GfxInternal
         shader_args.push_back(L"-I"); shader_args.push_back(L".");
         shader_args.push_back(L"-T"); shader_args.push_back(wshader_profile.data());
         shader_args.push_back(L"-HV 2021");
+        shader_args.push_back(L"-Wno-parameter-usage");
+        shader_args.push_back(L"-Wno-uninitialized");
+        shader_args.push_back(L"-Wno-conditional-uninitialized");
+        shader_args.push_back(L"-Wno-sometimes-uninitialized");
+        shader_args.push_back(L"-fdiagnostics-format=msvc");
```
gboisse, line 8891: Should these be added only when `debug_shaders_` is enabled? (assuming this is part of the workaround)

gfx.cpp
```diff
@@ -8921,7 +8928,7 @@ class GfxInternal
         if(debug_shaders_)
         {
             shader_args.push_back(DXC_ARG_DEBUG);
-            shader_args.push_back(DXC_ARG_OPTIMIZATION_LEVEL0);
+            shader_args.push_back(DXC_ARG_SKIP_OPTIMIZATIONS);
```
gboisse, line 8931: Out of curiosity, what's the difference between `DXC_ARG_OPTIMIZATION_LEVEL0` and `DXC_ARG_SKIP_OPTIMIZATIONS`?

gfx_core.h
gboisse: hehe, not a fan of these dynamic allocations but that's alright 🙂
`printf()` is already so slow in the 1st place anyway...

gfx.cpp
maoliver-amd: O0 performs minor optimisations and Od doesnt do anything. Od can result in some terribly performing code compilation compared to O0 but O0 can also optimise away some variables such that they cant be viewed in PIX

gfx.cpp
maoliver-amd: Possibly. I havnt actually tested whether they are needed for release builds as there is a lot of permutations to go through (they disable diagnostics so I figured thees no harm).
These are all inherited from the dxc-6.8 branch though and the only new one added for this PR is `shader_args.push_back(L"-fdiagnostics-format=msvc");` which is required all the time to get the compiler output in the correct format

gfx_core.h
maoliver-amd: Yeah, unfortunately shader compilation output can be really big depending on how many errors/warnings are generated. So I didnt see any other way around it. It is only in the debug/error path so shouldnt have any runtime performance implications when everything is working correctly

gfx.cpp
gboisse: Sounds good, let's keep it as is then.

gfx_core.h
gboisse: Yep, that's correct.

gfx.cpp
maoliver-amd: Unfortunately microsoft doesnt really provide much information on what exactly is difference ([https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/d3dcompile-constants](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/d3dcompile-constants))
Od is less likely to cause compiler crashes though in my testing

PR #170, opened 2025-06-03
Fix errors from debug layers when using DXR-1.0 and SBTs

gboisse:


gfx.cpp
```diff
@@ -6812,6 +6812,7 @@ class GfxInternal
                             }
                         }
                     }
+                    transitionResource(sbt_buffer, D3D12_RESOURCE_STATE_NON_PIXEL_SHADER_RESOURCE);
                     command_list_->CopyBufferRegion(sbt_buffer.resource_, dst_offset, upload_buffer.resource_,
                         src_offset, upload_buffer_offset - src_offset);
```
gboisse, line 6817: Doesn't look like transitioning to `D3D12_RESOURCE_STATE_COPY_DEST` is required here (or at least the debug layer doesn't complain about it).

maoliver-amd:
Ive looked into this and the error is due to implicit state transitions changing the sbt buffer from `D3D12_RESOURCE_STATE_COMMON` to `D3D12_RESOURCE_STATE_COPY_DEST` by the `CopyBufferRegion` on line 6815. Since theres no `submitPipelineBarriers()` before that copy function then the state transitions get processed later on in execution where the resource state no longer matches.

See [https://learn.microsoft.com/en-us/windows/win32/direct3d12/using-resource-barriers-to-synchronize-resource-states-in-direct3d-12#common-state-promotion](https://learn.microsoft.com/en-us/windows/win32/direct3d12/using-resource-barriers-to-synchronize-resource-states-in-direct3d-12#common-state-promotion)

Ive created an alternate PR that corrects this

PR #171, opened 2025-06-04
Fix transition state debug errors for sbt buffers.

maoliver-amd:
Alternate fix for #170 

The error is due to implicit state transitions changing the sbt buffer from D3D12_RESOURCE_STATE_COMMON to D3D12_RESOURCE_STATE_COPY_DEST by the CopyBufferRegion on line 6815. Since theres no submitPipelineBarriers() before that copy function then the state transitions get processed later on in execution where the resource state no longer matches.

See https://learn.microsoft.com/en-us/windows/win32/direct3d12/using-resource-barriers-to-synchronize-resource-states-in-direct3d-12#common-state-promotion

This PR correctly transitions the resource to `D3D12_RESOURCE_STATE_COPY_DEST` with a pipeline submission so ensure proper barrier placement.

As a future update `transitionResource` should be updated so that it does not submit barriers when an implicit state transition can be used instead. This would provide fewer barriers and improve performance

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -6761,6 +6761,20 @@ private:
             sbt->kernel_ = bound_kernel_; // update bound kernel
             for(uint32_t i = 0; i < kGfxShaderGroupType_Count; ++i)
             {
+                Buffer &sbt_buffer = buffers_[sbt->sbt_buffers_[i]];
+                bool    transition = false;
+                for(auto &shader_record : sbt->shader_records_[i])
+                {
+                    Sbt::ShaderRecord &sbt_record = shader_record.second;
+                    if(sbt_record.id_ == sbt_record.commited_id_ && !invalidate_sbt_descriptors && !invalidate_sbt_parameters) continue;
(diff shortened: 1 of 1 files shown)


gfx.cpp
```diff
@@ -6812,11 +6812,14 @@ class GfxInternal
                             }
                         }
                     }
+                    transitionResource(sbt_buffer, D3D12_RESOURCE_STATE_COPY_DEST);
+                    submitPipelineBarriers();
```
gboisse, line 6816: Issue with this is that we know have this `submitPipelineBarriers()` happening in a loop, some minimal batching would be good (like pre-transition all SBT buffers before the copies, then transition back after the copies, rather than one by one like this).

maoliver-amd: Good point, I was initially thinking that if we auto handle implicit transitions then this would be a noop. But ive updated the PR to batch the transitions anyway

gboisse (approved):

PR #173, opened 2025-06-06
Add support for detecting and using implicit resource transitions.

maoliver-amd:
D3D12 can implicitly promote resource states under certain circumstances without needing an explicit resource barrier. This is a potential source of performance improvement as it reduces the number of barriers and subsequent pipeline submits that are required to transition resources

[https://learn.microsoft.com/en-us/windows/win32/direct3d12/using-resource-barriers-to-synchronize-resource-states-in-direct3d-12#implicit-state-transitions](https://learn.microsoft.com/en-us/windows/win32/direct3d12/using-resource-barriers-to-synchronize-resource-states-in-direct3d-12#implicit-state-transitions)

This also exposed a few potential bugs in gfx's existing implementation where the internal stored resource state was not matching the actual resource state due to implicit transitions. This should now be corrected by identifying these transitions accordingly.

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -103,8 +103,8 @@ class GfxInternal
     ID3D12GraphicsCommandList4 *dxr_command_list_ = nullptr;
     ID3D12GraphicsCommandList6 *mesh_command_list_ = nullptr;
     ID3D12CommandAllocator **command_allocators_ = nullptr;
-    std::vector<IAmdExtD3DDevice1 *> amd_ext_devices_;
     ID3D12DebugCommandList *dbg_command_list_ = nullptr;
+    std::vector<IAmdExtD3DDevice1 *> amd_ext_devices_;
 
     HANDLE fence_event_ = {};
     uint32_t fence_index_ = 0;
@@ -403,12 +403,12 @@ class GfxInternal
 
         void *data_ = nullptr;
         uint64_t data_offset_ = 0;
+        bool *transitioned_ = nullptr;
         ID3D12Resource *resource_ = nullptr;
         uint32_t *reference_count_ = nullptr;
         D3D12MA::Allocation *allocation_ = nullptr;
         D3D12_RESOURCE_STATES *resource_state_ = nullptr;
         D3D12_RESOURCE_STATES initial_resource_state_ = D3D12_RESOURCE_STATE_COMMON;
-        bool *transitioned_ = nullptr;
     };
     GfxArray<Buffer> buffers_;
     GfxHandles buffer_handles_;
@@ -423,6 +423,7 @@ class GfxInternal
         inline bool isInterop() const { return (allocation_ == nullptr ? true : false); }
 
         uint32_t flags_ = 0;
+        bool transitioned_ = false;
         float clear_value_[4] = {};
         ID3D12Resource *resource_ = nullptr;
         D3D12MA::Allocation *allocation_ = nullptr;
@@ -430,7 +431,6 @@ class GfxInternal
         std::vector<uint32_t> rtv_descriptor_slots_[D3D12_REQ_MIP_LEVELS];
         D3D12_RESOURCE_STATES resource_state_ = D3D12_RESOURCE_STATE_COMMON;
         D3D12_RESOURCE_STATES initial_resource_state_ = D3D12_RESOURCE_STATE_COMMON;
-        bool transitioned_ = false;
     };
     GfxArray<Texture> textures_;
     GfxHandles texture_handles_;
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -3224,8 +3224,8 @@ public:
         if(gfx_dst.resource_ == gfx_src.resource_)
             return GFX_SET_ERROR(kGfxResult_InvalidOperation, "Cannot copy between two buffer objects that are pointing at the same resource; use an intermediate buffer");
         bool transitions = false;
-        if(dst.cpu_access == kGfxCpuAccess_None) transitions = transitionResource(gfx_dst, D3D12_RESOURCE_STATE_COPY_DEST, true);
-        if(src.cpu_access == kGfxCpuAccess_None) transitions |= transitionResource(gfx_src, D3D12_RESOURCE_STATE_COPY_SOURCE, true);
+        if(dst.cpu_access == kGfxCpuAccess_None) transitions = transitionResource(gfx_dst, D3D12_RESOURCE_STATE_COPY_DEST, kTransitionType_Implicit);
+        if(src.cpu_access == kGfxCpuAccess_None) transitions |= transitionResource(gfx_src, D3D12_RESOURCE_STATE_COPY_SOURCE, kTransitionType_Implicit);
         if(transitions)
             submitPipelineBarriers();   // transition our resources if needed
         command_list_->CopyBufferRegion(gfx_dst.resource_, gfx_dst.data_offset_,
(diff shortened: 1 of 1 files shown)


gfx.cpp
```diff
@@ -104,6 +104,7 @@ class GfxInternal
     ID3D12GraphicsCommandList6 *mesh_command_list_ = nullptr;
     ID3D12CommandAllocator **command_allocators_ = nullptr;
     std::vector<IAmdExtD3DDevice1 *> amd_ext_devices_;
+    ID3D12DebugCommandList *dbgCommandList = nullptr;
```
maoliver-amd, line 107: Only used for debugging that resource states match

gfx.cpp
```diff
@@ -1963,7 +1975,7 @@ class GfxInternal
         width = GFX_MAX(width, 1u);
         height = GFX_MAX(height, 1u);
         mip_levels = GFX_MIN(GFX_MAX(mip_levels, 1u), gfxCalculateMipCount(width, height));
-        D3D12_RESOURCE_STATES const resource_state = D3D12_RESOURCE_STATE_COPY_DEST;
+        D3D12_RESOURCE_STATES const resource_state = D3D12_RESOURCE_STATE_COMMON;
```
maoliver-amd, line 1978: All resources are created in COMMON state. The previous behaviour was actually a bug as GPU buffers where generally used as the dest for a copy operation when loading data into them which was previously implicitly setting the state to COPY_DEST. However if this operations was not done then this initial state would be incorrect

gfx.cpp
```diff
@@ -3202,9 +3214,11 @@ class GfxInternal
...
         if(gfx_dst.resource_ == gfx_src.resource_)
             return GFX_SET_ERROR(kGfxResult_InvalidOperation, "Cannot copy between two buffer objects that are pointing at the same resource; use an intermediate buffer");
-        if(dst.cpu_access == kGfxCpuAccess_None) transitionResource(gfx_dst, D3D12_RESOURCE_STATE_COPY_DEST);
-        if(src.cpu_access == kGfxCpuAccess_None) transitionResource(gfx_src, D3D12_RESOURCE_STATE_COPY_SOURCE);
-        submitPipelineBarriers();   // transition our resources if needed
+        bool transitions = false;
+        if(dst.cpu_access == kGfxCpuAccess_None) transitions = transitionResource(gfx_dst, D3D12_RESOURCE_STATE_COPY_DEST, true);
+        if(src.cpu_access == kGfxCpuAccess_None) transitions |= transitionResource(gfx_src, D3D12_RESOURCE_STATE_COPY_SOURCE, true);
+        if(transitions)
+            submitPipelineBarriers();   // transition our resources if needed
```
maoliver-amd, line 3221: `transitionResource` is now updated to return a bool signalling if a transition is actually required. If not then no pipeline submission is required

gfx.cpp
```diff
@@ -4540,6 +4561,36 @@ class GfxInternal
...
+            {
+                if (!texture->transitioned_ && (texture->resource_state_ & D3D12_RESOURCE_STATE_ALL_SHADER_RESOURCE) == texture->resource_state_)
+                {
+                    texture->resource_state_ = D3D12_RESOURCE_STATE_COMMON;
+                }
+                if(dbgCommandList != nullptr)
+                    dbgCommandList->AssertResourceState(texture->resource_, D3D12_RESOURCE_BARRIER_ALL_SUBRESOURCES, texture->resource_state_);
+            }
+        }
+    }
```
maoliver-amd, line 4592: Implicitly transitioned resources will also implicitly decay after a executeList call. This function will now auto change the resource states accordingly.
`AssertResourceState` is used here to check that the resources match what the internal state assumes.

gfx.cpp
```diff
@@ -9339,12 +9392,12 @@ class GfxInternal
         return kGfxResult_NoError;
     }
 
-    void transitionResource(Buffer &buffer, D3D12_RESOURCE_STATES resource_state)
+    bool transitionResource(Buffer &buffer, D3D12_RESOURCE_STATES resource_state, bool implicit_allowed = false)
```
maoliver-amd, line 9395: `transitionResource` also now takes and optional parameter which can be used to force disable implicit transition barriers.

gfx.cpp
```diff
@@ -4540,6 +4561,41 @@ class GfxInternal
...
+            for(uint32_t i = 0; i < textures_.size(); ++i)
+            {
+                auto *texture = textures_.at(i);
+                if(texture != nullptr)
+                {
+                    dbgCommandList->AssertResourceState(texture->resource_, D3D12_RESOURCE_BARRIER_ALL_SUBRESOURCES, texture->resource_state_);
+                }
+            }
+        }
+    }
```
maoliver-amd, line 4597: Implicitly transitioned resources will also implicitly decay after a executeList call. This function will now auto change the resource states accordingly.
AssertResourceState is used here to check that the resources match what the internal state assumes. Its in its own loop to prevent performance hits when not debugging. This function will cause debug errors to be logged next time the command list is executed

gfx.cpp
gboisse: Small thing but naming convention is snake case with trailing underscore in this codebase 🙂

gfx.cpp
```diff
@@ -405,8 +406,10 @@ class GfxInternal
         ID3D12Resource *resource_ = nullptr;
         uint32_t *reference_count_ = nullptr;
         D3D12MA::Allocation *allocation_ = nullptr;
-        D3D12_RESOURCE_STATES *resource_state_ = nullptr;
+        D3D12_RESOURCE_STATES resource_state_ = (D3D12_RESOURCE_STATES)UINT_MAX;
```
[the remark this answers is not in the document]
maoliver-amd: Ahh, I see. I had added the extra `parent_` variable to handle this. But the pointer makes sense. Ill change that back

gfx.cpp
maoliver-amd: fixed

gfx.cpp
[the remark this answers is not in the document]
maoliver-amd: Changed back, and moved `transitioned` to also be a pointer for the same reasons

maoliver-amd:
I updated a few more locations where implicit barriers can be used and where some initial state was not being set correctly.

gfx.cpp
gboisse: Could we turn this `bool` into an `enum`?
Say like `kTransitionType_Implicit` vs. `kTransitionType_WhateverIsTheOther`.
I believe that'd make the rest of the code a bit more readable rather than seeing some `transitionResource(..., true);` 🙂

maoliver-amd: I can do that

gfx.cpp
```diff
@@ -9402,7 +9402,15 @@ class GfxInternal
...
-    bool transitionResource(Buffer &buffer, D3D12_RESOURCE_STATES resource_state, bool implicit_allowed = false)
+    enum TransitionType
+    {
+        kTransitionType_Implicit = 0,
+        kTransitionType_Explicit,
+
+        kTransitionType_Count
+    };
+
+    bool transitionResource(Buffer &buffer, D3D12_RESOURCE_STATES resource_state, TransitionType implicit_allowed = kTransitionType_Explicit)
```
gboisse, line 9413: We could possibly rename from `implicit_allowed` to `transition_type`.
As-is, we're essentially having `if(implicit_allowed) -> implicit is not allowed` 🙂

PR #174, opened 2025-06-27
Fix debug layer errors when running headless

gboisse:


maoliver-amd (approved):
DiscardResource is a really poorly documented function, it appears there might be some optimisations around it when clearing but that might be a can of worms. So this works fine

gboisse:
> DiscardResource is a really poorly documented function, it appears there might be some optimisations around it when clearing but that might be a can of worms. So this works fine

Agreed, I do see this as little more than keeping the debug layer silent.
There's some more info on this blog: https://asawicki.info/news_1724_initializing_dx12_textures_after_allocation_and_aliasing

PR #178, opened 2025-07-14
Fix OOR Access On Shader Hash Calculation

tengel-lux:


gfx.cpp
```diff
@@ -9230,7 +9230,9 @@ class GfxInternal
                     for(size_t i = 1; i < shader_args.size(); ++i)
                     {
                         char buffer[64] = {};
-                        wcstombs(buffer, shader_args[i], sizeof(buffer));
+                        std::size_t ret = wcstombs(buffer, shader_args[i], sizeof(buffer));
+                        if (ret > 0) // make sure we have a null terminator since Hash function expects it
```
tengel-lux, line 9234: wcstombs may not insert the null terminator if the string is larger than 64 characters in this case

gfx.cpp
```diff
@@ -9230,7 +9230,8 @@ class GfxInternal
                     for(size_t i = 1; i < shader_args.size(); ++i)
                     {
                         char buffer[64] = {};
-                        wcstombs(buffer, shader_args[i], sizeof(buffer));
+                        std::size_t ret = wcstombs(buffer, shader_args[i], sizeof(buffer));
+                        buffer[GFX_MIN(ret, std::size(buffer) - 1)] = '\0'; // make sure we have a null terminator since Hash function expects it
```
tengel-lux, line 9234: `wcstombs` may not insert the null terminator if the string is larger than 64 characters in this case. We actually have defines larger than this so it was resulting in random hash values across different calls with the same input.

PR #179, opened 2025-07-23
Add DPI awareness

maoliver-amd:
This makes gfxWindow and gfxImgui DPI aware.

The main changes are that any application using gfxWindow will now set itself to be DPI aware and bypass windows inbuilt scaling. This results in 1:1 pixel output with windows scale artifacts on systems with scaling set to != 100%.
gfxImgui has been updated to detect DPI scaling and automatically apply the correct scale to UI elements and will detect and adapt to DPI changes at runtime.
gfxWindow will now attempt to preserve back buffer resolution across DPI changes so that if the user requests 1920x1080 then a 1920x1080 buffer will always be provided even when changing DPI at runtime. This differs from previous functionality where windows would scale the window based on the change in DPI

### DIFF
--- a/gfx_window.cpp
+++ b/gfx_window.cpp
@@ -193,6 +193,12 @@ public:
         callback_data_ = nullptr;
     }
 
+    inline float getDPIScale() const
+    {
+        UINT dpi = GetDpiForWindow(window_);
+        return (float)dpi / (float)USER_DEFAULT_SCREEN_DPI;
+    }
+
     static inline GfxWindowInternal *GetGfxWindow(GfxWindow window) { return reinterpret_cast<GfxWindowInternal *>(window.handle); }
 
 private:
@@ -378,4 +384,11 @@ bool gfxWindowUnregisterDropCallback(GfxWindow window)
     if(!gfx_window) return false; // invalid window handle
     gfx_window->unregisterDropCallback();
     return true;
+}
+
+float gfxWindowGetDPIScale(GfxWindow window)
+{
+    GfxWindowInternal *gfx_window = GfxWindowInternal::GetGfxWindow(window);
+    if(!gfx_window) return 1.0f; // invalid window handle
+    return gfx_window->getDPIScale();
 }
--- a/gfx_window.cpp
+++ b/gfx_window.cpp
@@ -46,6 +46,9 @@ class GfxWindowInternal
     bool is_previous_key_down_[VK_OEM_CLEAR] = {};
     void (*drop_callback_)(char const *, uint32_t, void *) = nullptr;
     void *callback_data_ = nullptr;
+    GfxCreateWindowFlags flags_;
+    uint32_t window_width_;
+    uint32_t window_height_;
 
 public:
     GfxWindowInternal(GfxWindow &window) { window.handle = reinterpret_cast<uint64_t>(this); }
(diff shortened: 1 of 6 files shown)


gfx_window.cpp
```diff
@@ -115,12 +121,20 @@ class GfxWindowInternal
...
             }
         }
+        else
+        {
+            RECT window_rect = { 0, 0, (LONG)window_width_, (LONG)window_height_ };
+            UINT dpi = GetDpiForWindow(window_);
+            AdjustWindowRectExForDpi(&window_rect, window_style, FALSE, window_style_ex, dpi);
+            SetWindowPos(window_, NULL, 0, 0, window_rect.right - window_rect.left,
+                window_rect.bottom - window_rect.top, SWP_NOZORDER | SWP_NOACTIVATE | SWP_NOMOVE);
+        }
```
maoliver-amd, line 131: `AdjustWindowRect` is now DPI aware. This requires moving the window sizing till after window creation as you need a valid window to get the associated monitors DPI scaling.

gfx_window.cpp
```diff
@@ -228,12 +248,15 @@ class GfxWindowInternal
...
             switch(message)
             {
             case WM_SIZE:
                 {
-                    gfx_window->is_minimized_ = IsIconic(gfx_window->window_);
-                    gfx_window->is_maximized_ = IsZoomed(gfx_window->window_);
+                    gfx_window->is_minimized_  = IsIconic(gfx_window->window_);
+                    gfx_window->is_maximized_  = IsZoomed(gfx_window->window_);
+                    gfx_window->window_width_  = LOWORD(l_param);
+                    gfx_window->window_height_ = HIWORD(l_param);
```
maoliver-amd, line 259: User resizing is also now stored to allow it to be used across DPI rescales

gfx_window.cpp
```diff
@@ -266,6 +289,28 @@ class GfxWindowInternal
...
+                    RECT window_rect = { 0, 0, (LONG)gfx_window->window_width_, (LONG)gfx_window->window_height_ };
+                    DWORD const window_style = WS_OVERLAPPEDWINDOW & ~((gfx_window->flags_ & kGfxCreateWindowFlag_NoResizeWindow) != 0 ?
+                                                                           WS_THICKFRAME | WS_MINIMIZEBOX | WS_MAXIMIZEBOX : 0);
+                    DWORD const window_style_ex = WS_EX_OVERLAPPEDWINDOW & ((gfx_window->flags_ & kGfxCreateWindowFlag_AcceptDrop) != 0 ? WS_EX_ACCEPTFILES : 0);
+                    AdjustWindowRectExForDpi(&window_rect, window_style, FALSE, window_style_ex, dpi);
+                    SIZE* const out = (SIZE*)l_param;
+                    out->cx = window_rect.right - window_rect.left;
+                    out->cy = window_rect.bottom - window_rect.top;
+                    return TRUE;
+                }
```
maoliver-amd, line 304: `WM_GETDPISCALEDSIZE` is sent before a window is resized by windows based on a DPI change. If left unhandled then windows will automatically resize the window based on the DPI change (i.e. if changing from 100% to 150% scale then the window size will also be increased to 150%)
With this event it is possible to intersect the window resize and specify the desired size. In this case we maintain the user set width/height essentially preventing windows from resizing the back bufer. This way a back buffer of 1920x1080 for instance will stay at the exact size even when changing DPI

gboisse (approved):

PR #181, opened 2025-08-15
Fix fullscreen windows and gfxWindowToggleFullscreen

maoliver-amd:
This fixes setting fullscreen window values as well as adds a new function that allows a user to runtime toggle between fullscreen and windowed

Edit this also fixes a bug where drag and drop no longer worked correctly.

### DIFF
--- a/gfx_window.cpp
+++ b/gfx_window.cpp
@@ -75,9 +75,8 @@ public:
 
         RegisterClassEx(&window_class);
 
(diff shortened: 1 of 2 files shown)


gfx_window.cpp
```diff
@@ -49,6 +49,8 @@ class GfxWindowInternal
     GfxCreateWindowFlags flags_;
     uint32_t window_width_;
     uint32_t window_height_;
+    uint32_t original_window_width_;
+    uint32_t original_window_height_;
```
maoliver-amd, line 53: Needed for toggling between fullscreen and windowed

PR #182, opened 2025-08-26
Adds functions for explicitly controlling buffer/texture resource flags/state.

maoliver-amd:
Adds 2 new functions that allow a user to explicitly change a resources state. These functions are intended for external interop where shared resources need to be in a specific state.

Texture creation functions have also been extended to allow resource flags to be specified during creation. This reduces runtime cost by avoiding the need to create and copy textures to a new resource with the required flags.

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -1982,10 +1982,10 @@ public:
 
     GfxTexture createTexture2D(DXGI_FORMAT format, float const *clear_value)
     {
-        return createTexture2D(window_width_, window_height_, format, 1, clear_value, Texture::kFlag_AutoResize);
+        return createTexture2D(window_width_, window_height_, format, 1, clear_value, D3D12_RESOURCE_FLAG_NONE, Texture::kFlag_AutoResize);
     }
 
-    GfxTexture createTexture2D(uint32_t width, uint32_t height, DXGI_FORMAT format, uint32_t mip_levels, float const *clear_value, uint32_t flags = 0)
+    GfxTexture createTexture2D(uint32_t width, uint32_t height, DXGI_FORMAT format, uint32_t mip_levels, float const *clear_value, D3D12_RESOURCE_FLAGS resource_flags, uint32_t flags = 0)
     {
         GfxTexture texture = {};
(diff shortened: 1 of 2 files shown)


gfx.cpp
```diff
@@ -4976,6 +5093,50 @@ class GfxInternal
...
+                for(uint32_t j = 0; j < depth; ++j)
+                    ensureTextureHasDepthStencilView(texture, gfx_texture, i, j);
+                if(texture.is3D())
+                    depth = GFX_MAX(depth >> 1, 1u);
+            }
+        }
+        if(transitionResource(gfx_texture, state))
+            submitPipelineBarriers();
+        return kGfxResult_NoError;
+    }
```
maoliver-amd, line 5138: This uses the same code as used elsewhere to ensure that textures have the correct usage flag to allow the requested resource state to be transitioned to. `ensureTextureHasRenderTargetView` and `ensureTextureHasDepthStencilView` both call `ensureTextureHasUsageFlag` to ensure correct flags. These are the only flags that require handling within gfx.

gfx.cpp
```diff
@@ -2037,10 +2038,38 @@ class GfxInternal
...
+                for(size_t j = 0; j < gfx_texture.dsv_descriptor_slots_[i].size(); ++j)
+                    gfx_texture.dsv_descriptor_slots_[i][j] = 0xFFFFFFFFu;
+            }
+            for(uint32_t i = 0; i < ARRAYSIZE(gfx_texture.rtv_descriptor_slots_); ++i)
+            {
+                gfx_texture.rtv_descriptor_slots_[i].resize(resource_desc.DepthOrArraySize);
+                for(size_t j = 0; j < gfx_texture.rtv_descriptor_slots_[i].size(); ++j)
+                    gfx_texture.rtv_descriptor_slots_[i][j] = 0xFFFFFFFFu;
+            }
+        }
```
maoliver-amd, line 2068: When creating new textures using input resource flags the texture has to be transitioned to the corresponding resource state and the descriptor slots have to be initialised to match code in `ensureTextureHasUsageFlag`

gboisse (approved):

PR #183, opened 2025-09-29
Update DirectX Agility and DXC to latest versions

maoliver-amd:
This fixes compilation with latest agility SDK while also updating the cmake versions.

TinyEXR was also updates to latests while I was at it.


maoliver-amd:
Ive been using the latests agility and DXC for several months now without issue.

PR #184, opened 2025-10-03
Clean old cache/pdb files

maoliver-amd:
This adds the ability to detect when multiple old shader cache and pdb files have been created for a shader kernel and removes any older ones. This helps prevent file bloat from continually creating new cache/pdbs

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -9372,8 +9372,12 @@ private:
         }
 
         uint64_t shader_key = 0;
+        std::wstring shader_key_dir;
         std::wstring shader_key_bytecode;
         std::wstring shader_key_reflection;
+        std::wstring const shader_cache_dir = L"./shader_cache/";
+        std::wstring const shader_pdb_dir = L"./shader_pdb/";
+        constexpr uint32_t max_cached_files = 5;
         if constexpr(std::is_same<ID3D12ShaderReflection, REFLECTION_TYPE>::value)
         {
             IDxcResult *dxc_preprocess = nullptr;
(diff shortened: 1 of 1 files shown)


gfx.cpp
```diff
@@ -9503,25 +9529,38 @@ class GfxInternal
...
-            std::string_view const shader_pdb_directory = "./shader_pdb";
-            std::wstring const wpdb_name(dxc_pdb_name->GetStringPointer(), dxc_pdb_name->GetStringLength());
-            std::vector<char> pdb_name(wcstombs(nullptr, wpdb_name.c_str(), 0) + 1);
-            wcstombs(pdb_name.data(), wpdb_name.c_str(), pdb_name.size());
-            shader_file.resize(shader_pdb_directory.size() + pdb_name.size() + 1);
-            GFX_SNPRINTF(shader_file.data(), shader_file.size(), "%s/%s", shader_pdb_directory.data(), pdb_name.data());
             if(!created_shader_pdb_directory)
             {
-                int32_t const result = _mkdir(shader_pdb_directory.data());
+                int32_t const result = _wmkdir(shader_pdb_dir.c_str());
```
maoliver-amd, line 9534: All new shader cache and pdb file paths are now wstring by default. This saves a conversion and also supports unicode code paths for output paths/filenames

gfx.cpp
```diff
@@ -9372,8 +9372,12 @@ class GfxInternal
...
 
         uint64_t shader_key = 0;
-        std::string shader_key_bytecode;
-        std::string shader_key_reflection;
+        std::wstring shader_key_dir;
+        std::wstring shader_key_bytecode;
+        std::wstring shader_key_reflection;
+        std::wstring const shader_cache_dir = L"./shader_cache/";
+        std::wstring const shader_pdb_dir = L"./shader_pdb/";
+        constexpr uint32_t max_cached_files = 5;
```
maoliver-amd, line 9380: The `max_cached_files` variable is used to set the maximum number of cache/pdb files stored for a kernel before they are deleted. It is default set at 5 so that when changing options it doesnt always trigger a recompile but minimises the potential number of redundant cache/pdbs

gfx.cpp
```diff
@@ -9372,8 +9372,12 @@ class GfxInternal
         }
 
         uint64_t shader_key = 0;
-        std::string shader_key_bytecode;
-        std::string shader_key_reflection;
+        std::wstring shader_key_dir;
+        std::wstring shader_key_bytecode;
+        std::wstring shader_key_reflection;
```
maoliver-amd, line 9377: cache/pdb files are now stored using `shader_key_dir` This is built from the source code filename and path aswell as kernel entry and type. This should create a relatively unique name kernel shaders and help limit hash collisions. Files are then stored in their respective directories in a subfolder named `shader_key_dir`. With the hash key as the filename (hash keys are now converted to string as hex instead of uint).
Using a subfolder reduces hash collisions and makes it faster to search for old files as it can be done on a per kernel basis by just checking the corresponding subfolder.
Putting pdbs in a subfolder does not effect the ability for them to be used in debugging as pix will find them in subfolders automatically (tested and works fine)

gfx.cpp
```diff
@@ -9503,25 +9529,38 @@ class GfxInternal
...
+                if(is_regular_file(*iter))
+                    if(iter->path().extension() == ".pdb")
+                        pdb_files.emplace_back(*iter);
+            if(pdb_files.size() > max_cached_files)
+            {
+                std::sort(pdb_files.begin(), pdb_files.end(), [](const std::filesystem::path &a, const std::filesystem::path &b) {
+                    return last_write_time(a) > last_write_time(b); });
+                for(size_t i = max_cached_files; i < pdb_files.size(); ++i)
+                    std::filesystem::remove(pdb_files[i]);
+            }
```
maoliver-amd, line 9563: Removing old files is done by iterating the kernel subfolder, collecting a list fo files with the correct extension, sorting based on modfied date, and then removing all files older then the first `max_cached_files`

PR #185, opened 2025-10-21
Changes for Frame Generation

tengel-lux:
* Allow replacing the swap chain by a custom one
* Allow ImGui to optionally be rendered to a texture instead of the backbuffer


gfx.cpp
```diff
@@ -1674,6 +1669,13 @@ class GfxInternal
         gfxFree(back_buffers_);
         back_buffers_ = nullptr;
 
+        if(swap_chain_ != nullptr) // Done after removal of attachments as they hold 1 ref
+        {
+            [[maybe_unused]] uint32_t count = swap_chain_->Release();
+            GFX_ASSERT(count == 0);
```
gboisse: What's the reason for asserting on `count` here?

tengel-lux: I added as a sanity check since we have more than one place that may create a swapchain now. And ofc I wanted to make sure they get released, and that this is the place it occurs. I can remove it though, was more for debugging anyways..

gfx.h
```diff
@@ -480,6 +480,9 @@ ID3D12GraphicsCommandList *gfxGetCommandList(GfxContext context);
 GfxResult gfxSetCommandList(GfxContext context, ID3D12GraphicsCommandList *command_list);
 GfxResult gfxResetCommandListState(GfxContext context); // call this function before returning to gfx after externally modifying the state on the command list
 
+IDXGISwapChain4* gfxGetSwapChain(GfxContext context);
+GfxResult gfxSetSwapChain(GfxContext context, IDXGISwapChain4 *swapchain);
+
```
maoliver-amd: probably should be put down with the other interop functions. Same with the other 3 new functions (and be nice and move the definitions to the new order as well)

gfx.cpp
```diff
@@ -5208,6 +5207,84 @@ class GfxInternal
...
+                back_buffers_[i]->Release();
+                back_buffers_[i] = nullptr;
+                freeRTVDescriptor(back_buffer_rtvs_[i]);
+            }
+            swap_chain_->Release();
+            swap_chain_ = nullptr;
+        }
+        sync();
+        // bind the new swap chain
+        if (swapchain)
```
maoliver-amd: and this one needs removal of the space

gfx.cpp
```diff
@@ -11107,6 +11185,41 @@ HANDLE gfxBufferCreateSharedHandle(GfxContext context, GfxBuffer buffer)
...
+{
+    GfxInternal *gfx = GfxInternal::GetGfx(context);
+    if(!gfx) return nullptr;
+    return gfx->getSwapChain();
+}
+
+GfxResult gfxSetSwapChain(GfxContext context, IDXGISwapChain4 *swapchain)
+{
+    GfxInternal *gfx = GfxInternal::GetGfx(context);
+    if (!gfx) return kGfxResult_InvalidParameter;
```
maoliver-amd: space between if again

gfx_imgui.cpp
```diff
@@ -462,11 +462,21 @@ class GfxImGuiInternal
...
+                "   output = mul(BT2020ToBT709, ST2084);\r\n"
+                "#endif\r\n"
+                "#if defined(OUTPUT_HDR)\r\n"
                 "    imgui.rgb *= ReferenceWhiteAdjust;\r\n"
                 "#endif\r\n"
-                "    output = lerp(imgui.rgb, output, imgui.a);\r\n"
-                "#ifdef OUTPUT_SRGB\r\n"
+                "output = lerp(imgui.rgb, output, imgui.a);\r\n"
+                "// Convert to the output space.\r\n"
+                "#if defined(OUTPUT_SRGB)\r\n"
```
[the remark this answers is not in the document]
tengel-lux: reverted

gfx.h
```diff
@@ -497,6 +499,14 @@ GfxResult gfxTextureSetResourceState(GfxContext context, GfxTexture texture, D3D
 
 HANDLE gfxBufferCreateSharedHandle(GfxContext context, GfxBuffer buffer);
 
+ID3D12Resource *gfxGetBackBuffer(GfxContext context);
+
+ID3D12RootSignature *gfxGetKernelRootSignature(GfxContext context, GfxKernel kernel);
+ID3D12PipelineState *gfxGetKernelPipelineState(GfxContext context, GfxKernel kernel);
```
gboisse: Minor thing, but these should be named `gfxKernelGetRootSignature()` and `gfxKernelGetPipelineState()` respectively.

gboisse (approved):
Please fix these 2 function names, otherwise this is good for merging 🙂

gfx.h
tengel-lux: makes sense, fixed

PR #187, opened 2025-10-29
Compile D3dMemoryAllocator to use agility SDK

maoliver-amd:
This modifies cmake so that the the d3dmemoryallocator gets compiled using the bundled agility sdk instead of the system d3d12 from the windows SDK. This matches compilation with gfx (which uses agility sdk) and enables newer features only provided by agility to be used by d3d12ma.

Also bumped glm to latest 1.0.2 version.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -244,7 +244,7 @@ if(GFX_ENABLE_SCENE)
     FetchContent_Declare(
         glm
         GIT_REPOSITORY https://github.com/g-truc/glm.git
-        GIT_TAG        1.0.1
+        GIT_TAG        1.0.2
         SOURCE_DIR     "${CMAKE_CURRENT_SOURCE_DIR}/third_party/glm/"
         FIND_PACKAGE_ARGS NAMES glm
     )


CMakeLists.txt
```diff
@@ -159,6 +142,24 @@ else()
     target_link_libraries(gfx PRIVATE Microsoft::DirectXShaderCompiler)
 endif()
 
+set(GFX_D3D12MA_PATH "${CMAKE_CURRENT_SOURCE_DIR}/third_party/D3D12MemoryAllocator/")
```
maoliver-amd: Had to move d3d12ma until after agility sdk has been pulled

CMakeLists.txt
```diff
@@ -159,6 +142,24 @@ else()
...
+set(D3D12MA_AGILITY_SDK_DIRECTORY "${GFX_AGILITY_PATH}" CACHE STRING "")
```
maoliver-amd: This is the key line that enables d3d12ma to pull in the agility sdk

PR #188, opened 2025-11-07
Improve imgui color space handling

maoliver-amd:
This corrects some color space conversion issues due to the way imgui expects colors to be handled (i.e. it expects them to be wrong).
Since it was not correctly possible to render imgui using HDR using a direct render to backbuffer as blend operations should be done in linear space (well imgui expect srgb) and not in ST2048. This allows imgui rendering to be split so it renders to a texture then has a second function that composites that texture to screen which auto handles conversion to the correct color space/eotf.

ImGui rendering now supports an optional input texture parameter that can be used to render the UI to inplace of the backbuffer

The downside is that now gfx will ensure that imgui rendering is always performed on an 8bit sRGB buffer to ensure correctness and will error out if one is not provided. This is a change from previous behaviour

### DIFF
--- a/gfx_imgui.cpp
+++ b/gfx_imgui.cpp
@@ -59,21 +59,22 @@ class GfxImGuiInternal
     GfxBuffer *vertex_buffers_ = nullptr;
     GfxProgram imgui_program_ = {};
     GfxKernel imgui_kernel_ = {};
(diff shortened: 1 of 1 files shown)


maoliver-amd:
Here are some screen shots showing the difference.
Before:
image

After:
image

PR #201, opened 2026-04-02
Borderless Window Option

tengel-amd:
Convenient mode when one wants to compare screenshot against dumped images at same resolution

### DIFF
--- a/gfx_window.cpp
+++ b/gfx_window.cpp
@@ -67,6 +67,7 @@ class GfxWindowInternal
     bool is_minimized_ = false;
     bool is_maximized_ = false;
     bool is_close_requested_ = false;
+    bool is_full_screen_ = false;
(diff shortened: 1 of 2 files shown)


gfx_window.cpp
```diff
@@ -308,6 +308,10 @@ class GfxWindowInternal
 
     inline void toggleFullscreen() const
     {
+        // Borderless windows do not participate in fullscreen toggling.
+        if((flags_ & kGfxCreateWindowFlag_BorderlessWindow) != 0)
+            return;
```
gboisse: > Borderless windows do not participate in fullscreen toggling.

Why is that?

tengel-amd: good question, I guess I was having a hard time detecting the current state based on the window flags (whether full screen was enabled or not). But adding a boolean on the side also works, so just did that :)

gboisse (approved):

PR #202, opened 2026-05-11
Relative Shader File Names

tengel-amd:


gfx.cpp
```diff
@@ -9366,8 +9366,26 @@ class GfxInternal
...
+        auto file_path_relative = file_path_full;
+        for (size_t i = 0; i < program.include_paths_.size(); ++i)
+        {
+            auto include_path =
+                std::filesystem::path(program.include_paths_[i].c_str()).lexically_normal();
+            auto path_relative = std::filesystem::relative(file_path_full, include_path);
+            if (!path_relative.empty())
+            {
+                file_path_relative = include_path / path_relative;
+                break;
```
maoliver-amd, line 9382: Should this be breaking once one is found? Dont you want to go through all include folders and select the one that is a parent to all the others (and the shader file)? As the furthest parent may not be the first in the include list

PR #205, opened 2026-05-27
Add gfxTextureCreateSharedHandle.

maoliver-amd:
As it says on the tin. This adds a new function that allows sharing texture resources with other external APIs

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -5214,6 +5214,73 @@ public:
         return handle;
     }
 
+    HANDLE createTextureSharedHandle(GfxTexture const &texture)
+    {
+        HANDLE handle = {};
+        if(!texture_handles_.has_handle(texture.handle))
+        {
+            if(!!texture)
+                GFX_PRINT_ERROR(kGfxResult_InvalidParameter, "Cannot create shared handle from invalid texture object");
+            return handle;  // invalid texture object
+        }
+        Texture &gfx_texture = textures_[texture];
+        if(gfx_texture.isInterop())
+        {
+            GFX_PRINT_ERROR(kGfxResult_InvalidOperation, "Cannot create shared handle from interop texture object");
+            return handle;  // invalid texture object type
+        }
+        D3D12_HEAP_FLAGS heap_flags = D3D12_HEAP_FLAG_NONE;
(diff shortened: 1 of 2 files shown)


gfx.cpp
```diff
@@ -5214,6 +5214,73 @@ class GfxInternal
...
+            }
+            for(uint32_t i = 0; i < ARRAYSIZE(gfx_texture.rtv_descriptor_slots_); ++i)
+            {
+                gfx_texture.rtv_descriptor_slots_[i].resize(resource_desc.DepthOrArraySize);
+                for(size_t j = 0; j < gfx_texture.rtv_descriptor_slots_[i].size(); ++j)
+                    gfx_texture.rtv_descriptor_slots_[i][j] = 0xFFFFFFFFu;
+            }
+        }
+        WindowsSecurityAttributes security_attributes;
+        if(HRESULT status = device_->CreateSharedHandle(gfx_texture.resource_, &security_attributes, GENERIC_ALL, nullptr, &handle); !SUCCEEDED(status))
```
gboisse: Looks good, but any reason for passing `nullptr` here rather than the name of the resource as we do with shared buffers?

maoliver-amd: The name is used so that you can lookup a shared resource through win32 APIs by name instead of needing to have the handle already (which is something we dont currently do). I found though this prevents usage of this function in an interop situation where the texture is resized. So in a situation where a user has called this function, got the handle and passed that to some other function (say a 3rd party lib) which now holds that handle, then the user resizes the texture which creates a new texture resource and requires getting an updated handle. However, calling this function will fail as theres already an existing handle with the texture name (the old texture handle still be held by the other function). So having the name means that you have to flush out and remove all references to the initial handle before you can create another one. This prevents pipelining as the renderer need to be flushed and thus you cant have the default 3 frames in flight.

So this may be an issue with the buffer version that might want to be changed?

maoliver-amd:
OK Ive also removed the use of name from `createBufferSharedHandle` as well as a fixup in changing `!((heap_flags & D3D12_HEAP_FLAG_SHARED) != 0)` to `(heap_flags & D3D12_HEAP_FLAG_SHARED) == 0` by removing the double negative

gboisse (approved):

PR #208, opened 2026-06-01
Allow multiple geometries in BLAS builds

VladimirMakeew:
This PR introduces new api function `gfxRaytracingPrimitiveBuildBatch` which allows to pack multiple `GfxRaytracingPrimitive` into a single BLAS during build.
In order to properly group meshes into BLASes a new object type was introduced - `GfxRenderInstance`. It stores all meshes that have the same global transform. `GfxInstance` was renamed to `GfxMeshInstance` for clarity.
New api usage example:
```c++
for (uint32_t i = 0U, instanceID = 0U; i < gfxSceneGetRenderInstanceCount(scene); ++i)
{
    GfxRenderInstance const &render_instance = gfxSceneGetRenderInstances(scene)[i];
    std::vector<GfxRaytracingPrimitiveBatchElement> batch;
    
    for (size_t j = 0U; j < render_instance.instances.size(); ++j)
    {
        GfxRef<GfxMeshInstance> const &mesh_instance = render_instance.instances[j];
        GfxRaytracingPrimitive primitive = gfxCreateRaytracingPrimitive(gfx, acceleration_structure);
        glm::mat4 const row_major_transform = transpose(render_instance.transform);
        gfxRaytracingPrimitiveSetTransform(gfx, primitive, &row_major_transform[0][0]);
        gfxRaytracingPrimitiveSetInstanceID(gfx, primitive, instanceID);
        
        GfxRaytracingPrimitiveBatchElement element = {};
        element.primitive     = primitive;
        element.index_buffer  = index_buffer;
        element.vertex_buffer = vertex_buffer;
        element.vertex_stride = 0U;
        element.flags         = isOpaque(mesh_instance->material) ? kGfxBuildRaytracingPrimitiveFlag_Opaque : 0U;
        
        batch.push_back(element);
    }
    
    if (!batch.empty())
    {
        gfxRaytracingPrimitiveBuildBatch(gfx, batch.data(), batch.size());
    }
    
    instanceID += static_cast<uint32_t>(render_instance.instances.size());
}

gfxAccelerationStructureUpdate(gfx, acceleration_structure);
```

### DIFF
--- a/gfx_scene.cpp
+++ b/gfx_scene.cpp
@@ -177,6 +177,8 @@ class GfxSceneInternal
     GfxArray<uint64_t> instance_refs_;
     GfxArray<GfxMetadata> instance_metadata_;
     GfxHandles instance_handles_;
+    
+    std::vector<std::vector<uint64_t>> instances_by_transform_;
 
     static GfxArray<GfxScene> scenes_;
     static GfxHandles scene_handles_;
@@ -521,6 +523,11 @@ public:
     }
 
     static inline GfxSceneInternal *GetGfxScene(GfxScene scene) { return reinterpret_cast<GfxSceneInternal *>(scene.handle); }
+
+    inline std::vector<std::vector<uint64_t>> getInstancesBySameTransform() const
+    {
+        return instances_by_transform_;
+    }
 
 private:
     static inline uint32_t GetObjectIndex(uint64_t object_handle)
@@ -1958,10 +1965,14 @@ private:
                 if(it != meshes.end())
                 {
                     instances.reserve((*it).second.size());
+                    std::vector<uint64_t> same_transform;
+                    same_transform.reserve((*it).second.size());
+                    
                     for(size_t i = 0; i < (*it).second.size(); ++i)
                     {
                         GfxRef<GfxInstance> instance_ref = gfxSceneCreateInstance(scene);
                         instances.push_back(instance_ref);
+                        same_transform.push_back(instance_ref);
                         GfxConstRef<GfxMesh> mesh = it->second[i].first;
                         instance_ref->mesh      = mesh;
                         instance_ref->material  = it->second[i].second;
@@ -1983,6 +1994,8 @@ private:
                         else
                             instance_metadata.asset_file = asset_file;
                     }
+                    
+                    instances_by_transform_.push_back(same_transform);
                 }
             }
             GfxRef<GfxCamera> camera;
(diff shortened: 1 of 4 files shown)


gfx_scene.cpp
```diff
@@ -3622,3 +3635,10 @@ bool gfxSceneSetInstanceMetadata(GfxScene scene, uint64_t instance_handle, GfxMe
     if(!gfx_scene) return false;    // invalid parameter
     return gfx_scene->setObjectMetadata<GfxInstance>(instance_handle, metadata);
 }
+
+std::vector<std::vector<uint64_t>> gfxSceneGetInstancesBySameTransform(GfxScene scene)
```
maoliver-amd: This seems extremely hazardous when it comes to animations or the like. Id recommend just changing the gfxScene ABI so that theres an additional level above `GfxInstance` called `GfxRenderInstance` or something (might even be worth renaming `GfxInstance` to `GfxMeshInstance` to avoid mistakes with un-updated old code). Remove `transform` from `GfxInstance` and add it to `GfxRenderInstance`
```
struct GfxRenderInstance
{
    std::vector<GfxMeshInstance> instances;
    glm::mat4 transform = glm::mat4(1.0f);
};
```
For importing '.obj' files then there would simply be 1 `GfxRenderInstance` for whole scene. But for GLTFs then each `GfxRenderInstance` would map directly to a collapsed node (basically gfxs internal `GltfNode` with all the parent nodes collapsed - which is already what gfx does to create existing `GfxInstance` instead it would only have to be performed once for the node instead of for each instance within that node)

VladimirMakeew: Done,
I've added `GfxRenderInstance` with api methods similar to other gfx scene objects. `GfxInstance` was renamed to `GfxMeshInstance`, corresponding api function names were updated as well.
GfxRenderInstance api has the following functions:
- `gfxSceneCreateRenderInstance`
- `gfxSceneDestroyRenderInstance`
- `gfxSceneDestroyAllRenderInstances`
- `gfxSceneGetRenderInstanceCount`
- `gfxSceneGetRenderInstances`
- `gfxSceneGetRenderInstance`
- `gfxSceneGetRenderInstanceHandle`
- `gfxSceneGetRenderInstanceMetadata`
- `gfxSceneSetRenderInstanceMetadata`

Template functions such as `gfxSceneGetObjectCount` are also supported.

GLTF scene loading now produces `GfxRenderInstance`s with child `GfxMeshInstance`s that share the same global transform. OBJ scene loading support isn't done yet, I'll fix it after polishing api usage on Capsaicin side.

gfx_scene.h
```diff
@@ -411,29 +411,48 @@ GfxMetadata const &gfxSceneGetMeshMetadata(GfxScene scene, uint64_t mesh_handle)
 bool gfxSceneSetMeshMetadata(GfxScene scene, uint64_t mesh_handle, GfxMetadata const &metadata);
 
 //!
-//! Instance object.
+//! Mesh instance object.
 //!
 
-struct GfxInstance
+struct GfxMeshInstance
```
gboisse, line 417: I would place the `GfxScene` changes into their own PR to better separate the two changes.

VladimirMakeew: Done, functionality is now split across two PRs:
1. [GfxRenderInstance](https://github.com/gboisse/gfx/pull/217)
2. [Multiple geometries in a single BLAS](https://github.com/gboisse/gfx/pull/218)

VladimirMakeew:
Functionality of this PR was split and merged

PR #209, opened 2026-06-02
Fix for latest d3d12ma and tight alignment.

maoliver-amd:
Latest 3d312 agility SDK includes a new `D3D12_RESOURCE_FLAG_USE_TIGHT_ALIGNMENT` flag. D3d12ma picks this up and enables it as default. This requires ensuring that `Alignment` is always set to 0 otherwise the d3d runtime will error

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -1340,7 +1340,8 @@ public:
 
         D3D12MA::ALLOCATOR_DESC
         allocator_desc          = {};
-        allocator_desc.Flags    = D3D12MA::ALLOCATOR_FLAG_SINGLETHREADED;
+        allocator_desc.Flags =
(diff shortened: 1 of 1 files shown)


gfx.cpp
```diff
@@ -1340,7 +1340,8 @@ class GfxInternal
 
         D3D12MA::ALLOCATOR_DESC
         allocator_desc          = {};
-        allocator_desc.Flags    = D3D12MA::ALLOCATOR_FLAG_SINGLETHREADED;
+        allocator_desc.Flags =
+            D3D12MA::ALLOCATOR_FLAG_SINGLETHREADED;
```
gboisse, line 1344: This doesn't seem to be an actual change?

gboisse (approved):

PR #210, opened 2026-06-04
Fix a debug assertion due to double free.

maoliver-amd:
This reorders collection so that dependent objects are released first: their `collect()` implementations call `destroyBuffer()`, which fetches Buffer entries out of `buffers_` and releases them. Doing this before the raw `buffers_` sweep prevents a use-after-free / double-Release on `Buffer::reference_count_`, `resource_`, and `allocation_`, since `collect(Buffer&)` frees `reference_count_` but does not erase the entry from `buffers_`.

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -5949,7 +5949,7 @@ private:
         garbage_collection_.push_back(std::move(garbage));
     }
 
-    void collect(Buffer &buffer)
+    void collect(Buffer const &buffer)
     {
         if(buffer.reference_count_ != nullptr)
         {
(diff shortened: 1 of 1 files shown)


gfx.cpp
```diff
@@ -5972,9 +5972,17 @@ class GfxInternal
...
         gfxFree(buffer.reference_count_);
         gfxFree(buffer.transitioned_);
+        // Defensive: ensure a stale Buffer slot cannot be re-collected.
+        buffer.resource_ = nullptr;
+        buffer.allocation_ = nullptr;
+        buffer.resource_state_ = nullptr;
+        buffer.reference_count_ = nullptr;
+        buffer.transitioned_ = nullptr;
+        buffer.data_ = nullptr;
+        buffer.data_offset_ = 0;
```
gboisse: mmh, not sure I really like this "defensive" setup; best would be to understand why we can get into an (invalid) re-collection state in the first place.

gboisse: Is re-ordering the `collect()` in the `terminate()` function enough to fix the double frees?
In which case, I'd keep that and remove the "defensive" null-ing's.

maoliver-amd: Yes the reorder is enough to fix the error, the defensive code was purely so that the error is easily caught when debugging. Ill remove it if you want

gboisse (approved):

PR #211, opened 2026-06-05
Added message callback

SShikhali:
Added support for user provided message callback. When available, all GFX messages are redirected to it instead of printing to the console

### DIFF
--- a/gfx_core.h
+++ b/gfx_core.h
@@ -54,6 +54,13 @@ enum GfxConstant
     kGfxConstant_NumBindlessSlots = 2048
 };
 
+enum GfxVerbosity
+{
+    kGfxVerbosity_Error = 0,
+    kGfxVerbosity_Warning,
+    kGfxVerbosity_Info,
+};
+
 //!
 //! Public macros.
 //!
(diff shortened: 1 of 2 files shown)


maoliver-amd (approved):

gboisse:
This is a welcome change, but I'd say that we'd probably want the callback override to be at the context level rather than global.
This would mean either getting rid of the `GFX_PRINTLN`, `GFX_ASSERTMSG`, etc. macros... or having them accept a context object as part of the input parameters.

Probably the second option is the least intrusive here.
Then we can supply the message callback options to the `gfxCreateContext()` function.

gfx_core.h
```diff
@@ -172,6 +181,9 @@ enum GfxConstant
                                         inline operator bool() const { return !!handle; }                                           \
                                         private: uint64_t handle; char name[kGfxConstant_MaxNameLength + 1]
 
+typedef void (*GfxMessageCallback)(char const *message, GfxVerbosity verbosity);
```
gboisse, line 184: Capital `G` as this is a type, not a function.

gboisse (approved):

PR #212, opened 2026-06-08
Removed limit on bindless slots

SShikhali:
This PR removed hardcoded limit on bindless slots

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -771,7 +771,6 @@ class GfxInternal
                         break;
                     }
                 }
-
                 return descriptor_count;
             }
         };
@@ -7054,9 +7053,7 @@ private:
                 if(parameter.parameter_ != nullptr && parameter.id_ != parameter.parameter_->id_)
                 {
                     freeDescriptor(parameter.descriptor_slot_);
-
-                    uint32_t const descriptor_count = parameter.getNumDescriptors();
-                    parameter.descriptor_slot_ = allocateDescriptor(descriptor_count);
+                    parameter.descriptor_slot_ = allocateDescriptor(parameter.getNumDescriptors());
                 }
                 break;
             }
@@ -7102,9 +7099,7 @@ private:
                                 if(parameter.parameter_ != nullptr && parameter.id_ != parameter.parameter_->id_)
                                 {
                                     freeDescriptor(parameter.descriptor_slot_);
-
-                                    uint32_t const descriptor_count = parameter.getNumDescriptors();
-                                    parameter.descriptor_slot_ = allocateDescriptor(descriptor_count);
+                                    parameter.descriptor_slot_ = allocateDescriptor(parameter.getNumDescriptors());
                                 }
                                 break;
                             }
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -767,7 +767,7 @@ class GfxInternal
                         descriptor_count = parameter_->data_.image_.texture_count;
                         break;
                     default:
-                        GFX_ASSERT(!"Unknown type");
+                        GFX_ASSERTMSG(0, "Unknown descriptor type");
                         break;
                     }
                 }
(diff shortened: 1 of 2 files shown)


maoliver-amd (approved):
This seems surprisingly simple. Ill leave it for @gboisse to properly look this one over

gfx.cpp
```diff
@@ -6445,13 +6468,14 @@ class GfxInternal
                     case Kernel::Parameter::kType_Texture3D:
                     case Kernel::Parameter::kType_RWTexture3D:
                     case Kernel::Parameter::kType_TextureCube:
-                        descriptor_range.NumDescriptors = kGfxConstant_NumBindlessSlots;
+                        descriptor_range.NumDescriptors = UINT_MAX;
```
gboisse, line 6470: ha, didn't realize it was valid to pass `UINT_MAX` here, would you know where to find this in the spec., think I'd feel better about it if I'd see it mentioned there 🙂

gfx.cpp
```diff
@@ -6445,13 +6468,14 @@ class GfxInternal
...
                         break;
                     default:
                         GFX_PRINT_ERROR(kGfxResult_InternalError, "Bindless is only supported for buffer and texture objets");
                         break;
                     }
                 kernel_parameter.descriptor_count_ = descriptor_range.NumDescriptors;
+                kernel_parameter.bindless_ = descriptor_range.NumDescriptors == UINT_MAX;
```
gboisse: Minor thing, but I'd place some parenthesis around the condition to make this line a bit more readable, i.e.:
`kernel_parameter.bindless_ = (descriptor_range.NumDescriptors == UINT_MAX);`

gfx.cpp
```diff
@@ -7030,8 +7054,11 @@ class GfxInternal
                 if(parameter.parameter_ != nullptr && parameter.id_ != parameter.parameter_->id_)
                 {
                     freeDescriptor(parameter.descriptor_slot_);
-                    GFX_ASSERT(parameter.descriptor_count_ > 0);
-                    parameter.descriptor_slot_ = allocateDescriptor(parameter.descriptor_count_);
+
+                    uint32_t const descriptor_count = parameter.getNumDescriptors();
+                    GFX_ASSERT(descriptor_count > 0);
```
gboisse: Can't `descriptor_count` be `0` now? if the user hasn't bound anything to the bindless slot for instance?

gboisse:
Very good change for making bindless support trivial, but I'd still be a bit worried about that `UINT_MAX` and would like to know what happens at the driver level.

Hopefully we're not actually allocating 4+ billion of something each time bindless is used...

gboisse:
ha found it, excellent 🙂
unbound

gfx.cpp
```diff
@@ -751,6 +752,28 @@ class GfxInternal
...
+                    switch (parameter_->type_)
+                    {
+                    case Program::Parameter::Type::kType_Buffer:
+                        descriptor_count = parameter_->data_.buffer_.buffer_count;
+                        break;
+                    case Program::Parameter::Type::kType_Image:
+                        descriptor_count = parameter_->data_.image_.texture_count;
+                        break;
+                    default:
+                        GFX_ASSERT(!"Unknown type");
```
gboisse: Use `GFX_ASSERTMSG(0, "my message");` instead

gboisse:
Happy to approve this PR but please provide the explanation w.r.t. `descriptor_count` possibly be `0` (as we currently have an assert in the code against that particular case).

gfx.cpp
SShikhali: "The number of descriptors in the range. Use -1 or UINT_MAX to specify an unbounded size"
https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_descriptor_range

SShikhali:
Ok, I'll check if it's valid

SShikhali:
@gboisse Seems like `descriptor_count` being `0` is valid.
When we call `allocateDescriptor(0)` or `allocate_slots(0)`, we get `UINT_MAX` which is later detected and fallback descriptors (`dummy_descriptors_`) are being used instead. 
I see no validation errors from DX12

gfx.cpp
SShikhali: Reply is here: https://github.com/gboisse/gfx/pull/212#issuecomment-4729227172

gfx_core.h
```diff
@@ -51,7 +51,6 @@ enum GfxConstant
     kGfxConstant_MaxRenderTarget  = 8,
     kGfxConstant_MaxAnisotropy    = 8,
     kGfxConstant_MaxNameLength    = 64,
-    kGfxConstant_NumBindlessSlots = 2048
```
gboisse, line 54: Minor nit, but remove one space before the `=` sign to better align the enum definition.

gboisse (approved):
Approved (please fix minor nit reported above).

This was surprisingly simpler than anticipated, which is always good 🙂

PR #213, opened 2026-06-09
BVH Compaction support

SShikhali:
Adds BVH compaction support.

It requires extra CPU-GPU sync, so it's not recommended for dynamic geometry.
We wait for `BuildRaytracingAccelerationStructure` to finish, readback the compacted size, allocate a new buffer, execute `CopyRaytracingAccelerationStructure` to copy BVH to a new compacted buffer, and delete the old one.

It seems like compaction saves ~100mb of VRAM on `Bistro` and `Toyshop` scenes

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -7499,7 +7499,7 @@ private:
     {
         constexpr size_t compact_size = sizeof(D3D12_RAYTRACING_ACCELERATION_STRUCTURE_POSTBUILD_INFO_COMPACTED_SIZE_DESC);
         GFX_ASSERT(gfx_raytracing_primitive.type_ == RaytracingPrimitive::kType_Triangles); // should never happen
-        if (update && (gfx_raytracing_primitive.triangles_.build_flags_ & kGfxBuildRaytracingPrimitiveFlag_Updateable) == 0)
+        if(update && (gfx_raytracing_primitive.triangles_.build_flags_ & kGfxBuildRaytracingPrimitiveFlag_Updateable) == 0)
             return GFX_SET_ERROR(kGfxResult_InvalidOperation, "Cannot update a non-updateable raytracing primitive object");
         if(gfx_raytracing_primitive.triangles_.index_stride_ != 0 && !buffer_handles_.has_handle(gfx_raytracing_primitive.triangles_.index_buffer_.handle))
             return GFX_SET_ERROR(kGfxResult_InvalidOperation, "Cannot update a raytracing primitive that's pointing to an invalid index buffer object");
@@ -7542,15 +7542,15 @@ private:
         blas_inputs.pGeometryDescs = &geometry_desc;
         if(update)
             blas_inputs.Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PERFORM_UPDATE;
-        if (allow_compaction)
+        if(allow_compaction)
             blas_inputs.Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_COMPACTION;
-        if ((gfx_raytracing_primitive.triangles_.build_flags_ & kGfxBuildRaytracingPrimitiveFlag_Updateable) != 0)
+        if((gfx_raytracing_primitive.triangles_.build_flags_ & kGfxBuildRaytracingPrimitiveFlag_Updateable) != 0)
             blas_inputs.Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_UPDATE;
-        if ((gfx_raytracing_primitive.triangles_.build_flags_ & kGfxBuildRaytracingPrimitiveFlag_FastTrace) != 0)
+        if((gfx_raytracing_primitive.triangles_.build_flags_ & kGfxBuildRaytracingPrimitiveFlag_FastTrace) != 0)
             blas_inputs.Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PREFER_FAST_TRACE;
-        else if ((gfx_raytracing_primitive.triangles_.build_flags_ & kGfxBuildRaytracingPrimitiveFlag_FastBuild) != 0)
+        else if((gfx_raytracing_primitive.triangles_.build_flags_ & kGfxBuildRaytracingPrimitiveFlag_FastBuild) != 0)
             blas_inputs.Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PREFER_FAST_BUILD;
-        if ((gfx_raytracing_primitive.triangles_.build_flags_ & kGfxBuildRaytracingPrimitiveFlag_MinMemory) != 0)
+        if((gfx_raytracing_primitive.triangles_.build_flags_ & kGfxBuildRaytracingPrimitiveFlag_MinMemory) != 0)
             blas_inputs.Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_MINIMIZE_MEMORY;
         D3D12_RAYTRACING_ACCELERATION_STRUCTURE_PREBUILD_INFO blas_info = {};
         dxr_device_->GetRaytracingAccelerationStructurePrebuildInfo(&blas_inputs, &blas_info);
@@ -7572,9 +7572,9 @@ private:
             if(!gfx_raytracing_primitive.triangles_.bvh_buffer_)
                 return GFX_SET_ERROR(kGfxResult_OutOfMemory, "Unable to create raytracing primitive buffer");
         }
-        if (allow_compaction)
+        if(allow_compaction)
         {
-            if (!gfx_raytracing_primitive.triangles_.bvh_compact_size_buffer_)
+            if(!gfx_raytracing_primitive.triangles_.bvh_compact_size_buffer_)
             {
                 gfx_raytracing_primitive.triangles_.bvh_compact_size_buffer_ = createBuffer(compact_size, nullptr, kGfxCpuAccess_None);
                 gfx_raytracing_primitive.triangles_.bvh_compact_size_readback_buffer_ = createBuffer(compact_size, nullptr, kGfxCpuAccess_Read);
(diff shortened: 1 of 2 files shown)


maoliver-amd:
This looks like a good opportunity to add in the other BVH build flags as compact doesnt make much sense when gfx is still building BVHs with the update flag set. To really get the best from compaction we also need to *not* set update and allow setting the fast trace flag. I made a branch some time ago that added those flags that you can merge into this [maoliver/rt-primitive](https://github.com/gboisse/gfx/compare/master...maoliver/rt-primitive)

SShikhali:
> This looks like a good opportunity to add in the other BVH build flags as compact doesnt make much sense when gfx is still building BVHs with the update flag set. To really get the best from compaction we also need to _not_ set update and allow setting the fast trace flag. I made a branch some time ago that added those flags that you can merge into this [maoliver/rt-primitive](https://github.com/gboisse/gfx/compare/master...maoliver/rt-primitive)

I can see that your branch doesn't use `D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PERFORM_UPDATE` anymore. Is it intentional?

maoliver-amd:
> I can see that your branch doesn't use `D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PERFORM_UPDATE` anymore. Is it intentional?

Nope, that looks like a mistake on my part

SShikhali:
> This looks like a good opportunity to add in the other BVH build flags as compact doesnt make much sense when gfx is still building BVHs with the update flag set. To really get the best from compaction we also need to _not_ set update and allow setting the fast trace flag. I made a branch some time ago that added those flags that you can merge into this [maoliver/rt-primitive](https://github.com/gboisse/gfx/compare/master...maoliver/rt-primitive)

Do you suggest preventing users from using both `kGfxBuildRaytracingPrimitiveFlag_Updateable` and `kGfxBuildRaytracingPrimitiveFlag_Compact` at the same time? I think having both of them is technically valid, but it just will be slow

maoliver-amd:
> Do you suggest preventing users from using both `kGfxBuildRaytracingPrimitiveFlag_Updateable` and `kGfxBuildRaytracingPrimitiveFlag_Compact` at the same time? I think having both of them is technically valid, but it just will be slow

Setting both is valid and should be allowed. My comment was that currently in gfx master `D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_UPDATE` is always set, but after this PR it should be user configurable

SShikhali:
> > Do you suggest preventing users from using both `kGfxBuildRaytracingPrimitiveFlag_Updateable` and `kGfxBuildRaytracingPrimitiveFlag_Compact` at the same time? I think having both of them is technically valid, but it just will be slow
> 
> Setting both is valid and should be allowed. My comment was that currently in gfx master `D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_UPDATE` is always set, but after this PR it should be user configurable

Pushed

gfx.cpp
```diff
@@ -7489,7 +7497,10 @@ class GfxInternal
 
     GfxResult buildRaytracingPrimitiveTriangles(GfxRaytracingPrimitive const &raytracing_primitive, RaytracingPrimitive &gfx_raytracing_primitive, bool update)
     {
+        constexpr size_t compact_size = sizeof(D3D12_RAYTRACING_ACCELERATION_STRUCTURE_POSTBUILD_INFO_COMPACTED_SIZE_DESC);
         GFX_ASSERT(gfx_raytracing_primitive.type_ == RaytracingPrimitive::kType_Triangles); // should never happen
+        if (update && (gfx_raytracing_primitive.triangles_.build_flags_ & kGfxBuildRaytracingPrimitiveFlag_Updateable) == 0)
```
maoliver-amd: gfx code style uses `if(` instead of `if (` (aka no space)

SShikhali: Fixed

maoliver-amd (approved):

gfx.cpp
```diff
@@ -7569,13 +7614,50 @@ class GfxInternal
...
+        }
+        dxr_command_list_->BuildRaytracingAccelerationStructure(&build_desc, allow_compaction ? 1 : 0, allow_compaction ? &postbuild_info : nullptr);
+        if(allow_compaction)
+        {
+            Buffer& dst = buffers_[gfx_raytracing_primitive.triangles_.bvh_compact_size_readback_buffer_];
+            Buffer& src = buffers_[gfx_raytracing_primitive.triangles_.bvh_compact_size_buffer_];
+            transitionResource(src, D3D12_RESOURCE_STATE_COPY_SOURCE, kTransitionType_Implicit);
+            submitPipelineBarriers();
+            dxr_command_list_->CopyBufferRegion(dst.resource_, dst.data_offset_, src.resource_, src.data_offset_, compact_size);
+            finish();
```
gboisse: We cannot have a `finish()` in here... this is going to destroy our already-poor BVH build/update performance (see https://github.com/gboisse/gfx/pull/208).

That synchronization call must happen outside and be done by the application, so that we can do, e.g.:
```
gfxRaytracingPrimitiveBuild(gfx, bvh1);
gfxRaytracingPrimitiveBuild(gfx, bvh2);
gfxRaytracingPrimitiveBuild(gfx, bvh3);
gfxRaytracingPrimitiveBuild(gfx, bvh4);

gfxFinish(); // only one call to finish, not 3

gfxRaytracingPrimitiveCompact(gfx, bvh2);
gfxRaytracingPrimitiveCompact(gfx, bvh3);
gfxRaytracingPrimitiveCompact(gfx, bvh4);
```

gboisse: I like the fact that the readback buffer is handled directly by `gfx` as this seems more convenient in terms of library usage.
But it makes the need for synchronizing quite implicit and not obvious at an API level...

Ideally, it'd be good if we can detect the case where the user tries to build and compact with no synchronization in the middle and return a helpful error message (like "Post-build information not available yet, etc." or something).

SShikhali: Added `gfxRaytracingPrimitiveCompact` https://github.com/gboisse/gfx/pull/213/commits/7b253d6bc39ffe5ef9e42cdfab945c6b9c48e6d7

`finish_calls_counter_` is used to detect whether `finish()` was called in between the build & compact operations. If you have other ideas to detect it, please, let me know.

maoliver-amd: Technically we dont need to wait for a finish call at all, we just need to wait some number of frames until the compaction values are ready. ~~~So there are situations where the BVH is built, but then used in rendering for a few frames until compaction can be called on it.~~~ As such waiting for finish here would be incorrect

SShikhali: I'm sorry, I don't fully understand what you mean here.

Do you suggest to omit `gfxFinish` call and after N frames, just replace `bvh` internally for a compacted one?

maoliver-amd: No, sorry, what i was referring to is that checking `finish_calls_counter_` is not the most accurate way of detecting when the buffer is ready as we dont have control over how much time may occur between `gfxFinish` calls (as in bvh build may be done well before 3 frames). Its a simple guarantee to check finish count but a fence would be more accurate, each compacted BVH when running compaction can either check the fence state and return error or block and wait. Upto @gboisee which he prefers

gboisse: Indeed, `gfxFinish()` isn't the only valid way to synchronize here.

I've put up a proposal for a setup that accounts for calling `gfxFinish()` at least once and/or `gfxFrame()` enough times: https://github.com/gboisse/gfx/pull/227

gfx.cpp
```diff
@@ -7487,9 +7510,91 @@ class GfxInternal
...
+		if(gfx_raytracing_primitive.procedural_.finish_calls_state >= finish_calls_counter_)
+			return GFX_SET_ERROR(kGfxResult_InvalidOperation, "Post-build information is not available yet. Please, call `gfxFinish()` and try again");
+		Buffer& dst = buffers_[gfx_raytracing_primitive.triangles_.bvh_compact_size_readback_buffer_];
+		D3D12_RAYTRACING_ACCELERATION_STRUCTURE_POSTBUILD_INFO_COMPACTED_SIZE_DESC compact_size_desc{};
+		memcpy(&compact_size_desc, dst.data_, compact_size);
+		const size_t old_size = gfx_raytracing_primitive.triangles_.bvh_buffer_.getSize();
+		if(compact_size_desc.CompactedSizeInBytes == 0 || compact_size_desc.CompactedSizeInBytes > old_size)
+		{
+			GFX_SET_ERROR(kGfxResult_InternalError, "Can't readback compacted AS size. Possible sync issue.");
+			return kGfxResult_InternalError; // Should we return an error here? Or do we just continue normally?
```
gboisse, line 7528: Returning an error seems preferable to silently failing I would say.

Let's place the return before the `GFX_SET_ERROR()` call.

gfx.cpp
```diff
@@ -7487,9 +7510,91 @@ class GfxInternal
...
+		if(gfx_raytracing_primitive.procedural_.finish_calls_state >= finish_calls_counter_)
+			return GFX_SET_ERROR(kGfxResult_InvalidOperation, "Post-build information is not available yet. Please, call `gfxFinish()` and try again");
+		Buffer& dst = buffers_[gfx_raytracing_primitive.procedural_.bvh_compact_size_readback_buffer_];
+		D3D12_RAYTRACING_ACCELERATION_STRUCTURE_POSTBUILD_INFO_COMPACTED_SIZE_DESC compact_size_desc{};
+		memcpy(&compact_size_desc, dst.data_, compact_size);
+		const size_t old_size = gfx_raytracing_primitive.procedural_.bvh_buffer_.getSize();
+		if(compact_size_desc.CompactedSizeInBytes == 0 || compact_size_desc.CompactedSizeInBytes > old_size)
+		{
+			GFX_SET_ERROR(kGfxResult_InternalError, "Can't readback compacted AS size. Possible sync issue.");
+			return kGfxResult_InternalError; // Should we return an error here? Or do we just continue normally?
```
gboisse, line 7558: Same as https://github.com/gboisse/gfx/pull/213/changes#r3751494569.

gfx.cpp
gboisse: @maoliver-amd's proposal is interesting though; to make the API easier to use, we could detect whether some sync. is needed (using the mechanism from https://github.com/gboisse/gfx/pull/227), but instead of returning an error, print out a simple warning and call `finish()` internally.

That way, the compact function works out of the box, but with a path to make its use efficient when dealing with many BVHs 🙂

SShikhali:
@maoliver-amd @gboisse, considering @VladimirMakeew's PR includes these changes, should this one be closed?

gboisse:
> @maoliver-amd @gboisse, considering @VladimirMakeew's PR includes these changes, should this one be closed?

Yes.

SShikhali:
https://github.com/gboisse/gfx/pull/219

PR #215, opened 2026-06-22
Fix use after free crash in terminate.

maoliver-amd:
This is another one where the buffer was being destroyed before the object using it. Reordering the order of operations fixes it.


gboisse (approved):
Damn 😕 hopefully there isn't too many more of these:p

PR #218, opened 2026-07-01
Allow multiple geometries in BLAS builds

VladimirMakeew:
Add new api function `gfxRaytracingPrimitiveBuildBatch` which allows to pack multiple `GfxRaytracingPrimitive` into a single BLAS during build.

PR #219, opened 2026-07-03
Separate BLAS's from Acceleration Structure for Independent Builds

VladimirMakeew:
This PR introduces new API where BLAS's are separated from TLAS.

## New API requirements:
- BLAS objects independent from TLAS;
- Independent BLAS builds;
- Ability to swap BLAS's - for caching and LOD;
- Improve BLAS build and update performance;
- Support BLAS build flags and compaction;
- Improve TLAS update performance.

## API design:
- New object type `GfxGeometry` for geometry inside BLAS:
  - Can represent triangles or procedural geometry (clustered geometry is not implemented yet);
  - Can be used in multiple BLAS's;
  - Uses gfx handle system (`GFX_INTERNAL_HANDLE`).

Usage example:
```c++
// Create an opaque triangle geometry from index and vertex buffers
GfxGeometry geometry = gfxCreateGeometryTriangles(context, index_buffer, vertex_buffer, sizeof(Vertex));
gfxGeometrySetOpaque(context, geometry, true);
// Update buffers
gfxGeometryTrianglesUpdate(context, geometry, new_index_buffer, new_vertex_buffer, sizeof(Vertex));
// When done
gfxDestroyGeometry(context, geometry);
```
---
- New object type `GfxBottomLevelAccelerationStructure` that represents BLAS itself:
  - Does not depend on TLAS, can be swapped between TLAS's;
  - `GfxGeometry`-s can be added, removed, iterated;
  - Users can add multiple geometries to a single BLAS in order to pack them. This is a simplification of an API proposed in #218;
  - Functions for build, update, compact a single BLAS;
  - Functions for fast build or update multiple BLAS's together in a batch;
  - Function to report size after build.

Usage example:
```c++
// Create BLAS, it does not depend on TLAS and not attached to anything yet
GfxBottomLevelAccelerationStructure blas = gfxCreateBottomLevelAccelerationStructure(context);
// Add multiple geometries of different types to the same BLAS:
GfxGeometry triangles = gfxCreateGeometryTriangles(context, index_buffer, vertex_buffer, sizeof(Vertex));
gfxBottomLevelAccelerationStructureAddGeometry(context, blas, triangles);
GfxGeometry procedural = gfxCreateGeometryProcedural(context, aabb_buffer, sizeof(AABB));
gfxBottomLevelAccelerationStructureAddGeometry(context, blas, procedural);
```
---
- New object type `GfxTopLevelAccelerationStructureInstance` that represents instance inside TLAS:
  - `GfxBottomLevelAccelerationStructure` should be added to it;
  - Handles transform matrix, instanceID, mask, contribution to hit group index.

- New object type `GfxTopLevelAccelerationStructure` that represents TLAS:
  - `GfxTopLevelAccelerationStructureInstance`-s can be added, removed, iterated;
  - Functions for build and update;
  - Function to report size after build;
  - Interop support.
---
## API usage example:
```c++
vector<GfxBottomLevelAccelerationStructure> blasToBuild;
blasToBuild.reserve(gfxSceneGetRenderInstanceCount(scene));
vector<GfxBuildBottomLevelASFlags> blasBuildFlags;
blasBuildFlags.reserve(blasToBuild.size());
vector<uint32_t> compactedBlasIndices;
compactedBlasIndices.reserve(blasToBuild.size());

for (uint32_t i = 0U, instanceID = 0U; i < gfxSceneGetRenderInstanceCount(scene); ++i)
{
    GfxRenderInstance const &renderInstance = gfxSceneGetRenderInstances(scene)[i];
    if (renderInstance.instances.empty())
        continue;

    uint32_t const instance_index = gfxSceneGetObjectHandle<GfxRenderInstance>(scene, i);
    if (instance_index >= as_instances_.size())
    {
        as_instances_.resize(static_cast<size_t>(instance_index) + 1U);
        blas_.resize(static_cast<size_t>(instance_index) + 1U);
    }

    // Create tlas instance
    GfxTopLevelAccelerationStructureInstance &tlas_instance = as_instances_[instance_index];
    tlas_instance = gfxCreateTopLevelAccelerationStructureInstance(context);
    gfxTopLevelAccelerationStructureAddInstance(context, acceleration_structure_, tlas_instance);
    // Set instance data
    mat4 const transform = transpose(renderInstance.transform);
    gfxTopLevelAccelerationStructureInstanceSetTransform(context, tlas_instance, &transform[0][0]);
    gfxTopLevelAccelerationStructureInstanceSetInstanceID(context, tlas_instance, instanceID);
    gfxTopLevelAccelerationStructureInstanceSetInstanceContributionToHitGroupIndex(context, tlas_instance, instanceID * sbt_stride);
    // Create blas
    GfxBottomLevelAccelerationStructure &blas = blas_[instance_index];
    blas = gfxCreateBottomLevelAccelerationStructure(context);
    blas.setName(std::format("BottomLevelAccelerationStructure_{:d}", instanceID).c_str());
    gfxTopLevelAccelerationStructureInstanceSetBottomLevelAccelerationStructure(context, tlas_instance, blas);
    blasToBuild.push_back(blas);

    for (size_t j = 0U; j < renderInstance.instances.size(); ++j)
    {
        uint32_t const gpu_instance_index = instanceID + static_cast<uint32_t>(j);
        GfxBuffer const index_buffer = gfxCreateBufferRange<uint32_t>(context, index_buffer_, index_offset_idx, index_count);
        GfxBuffer const vertex_buffer = gfxCreateBufferRange<GPU::Vertex>(context, vertex_buffer_, vertex_offset_idx, vertex_count);

        // Create geometry
        GfxGeometry &geometry = geometries_[gpu_instance_index];
        geometry = gfxCreateGeometryTriangles(context, index_buffer, vertex_buffer, sizeof(GPU::Vertex));
        // Add geometry to blas
        gfxBottomLevelAccelerationStructureAddGeometry(context, blas, geometry);
        gfxGeometrySetOpaque(context, geometry, isOpaque);

        GfxBuildBottomLevelASFlags buildFlags = kGfxBuildBottomLevelASFlag_None;
        if (isAnimated)
            buildFlags |= kGfxBuildBottomLevelASFlag_Updateable | kGfxBuildBottomLevelASFlag_FastBuild;
        else
            buildFlags |= kGfxBuildBottomLevelASFlag_Compact | kGfxBuildBottomLevelASFlag_FastTrace;

        blasBuildFlags.push_back(buildFlags);
    }

    if ((blasBuildFlags.back() & kGfxBuildBottomLevelASFlag_Compact) == kGfxBuildBottomLevelASFlag_Compact)
        compactedBlasIndices.push_back(instance_index);

    instanceID += static_cast<uint32_t>(renderInstance.instances.size());
}

if (!blasToBuild.empty())
{
    // Build all BLAS's in a single batch: fast build using shared scratch buffer
    gfxBottomLevelAccelerationStructureBatchBuild(context, blasToBuild.data(), blasBuildFlags.data(), static_cast<uint32_t>(blasToBuild.size()));
}

if (!compactedBlasIndices.empty())
{
    // Wait compaction to finish
    gfxFinish(context);
    for (uint32_t const &index : compactedBlasIndices)
        gfxBottomLevelAccelerationStructureCompact(context, blas_[index]);
}

gfxTopLevelAccelerationStructureBuild(context, acceleration_structure_);
```


gfx.h
```diff
@@ -194,63 +194,90 @@ GfxSamplerState gfxCreateSamplerState(GfxContext context, D3D12_FILTER filter, D
...
-    GfxBuffer vertex_buffer;
-    uint32_t vertex_stride = 0;
-    GfxBuildRaytracingPrimitiveFlags flags = 0;
-};
+GfxResult gfxBottomLevelAccelerationStructureBuild(GfxContext context, GfxBottomLevelAccelerationStructure blas, GfxBuildBottomLevelASFlags flags);
+GfxResult gfxBottomLevelAccelerationStructureUpdate(GfxContext context, GfxBottomLevelAccelerationStructure blas);
+GfxResult gfxBottomLevelAccelerationStructureCompact(GfxContext context, GfxBottomLevelAccelerationStructure blas);
+uint64_t gfxBottomLevelAccelerationStructureGetDataSize(GfxContext context, GfxBottomLevelAccelerationStructure blas); // in bytes
+
+GfxResult gfxBottomLevelAccelerationStructureBatchBuild(GfxContext context, GfxBottomLevelAccelerationStructure const *blases, GfxBuildBottomLevelASFlags const *flags, uint32_t batch_size);
```
tengel-amd: imo the single blas builder should be removed. This already handles 1-N so imo there is no need to duplicate the code/functions. Then we can rename it to simply `gfxBuildBottomLevelAccelerationStructures`

gfx.cpp
```diff
@@ -444,64 +444,69 @@ class GfxInternal
...
+        GfxBuffer bvh_buffer_ = {};
+        uint64_t bvh_data_size_ = 0;
+        uint64_t finish_calls_state_ = 0; // Used to detect whether `finish()` was called in between build & compact operations
+        GfxBuffer bvh_compact_size_buffer_ = {};
+        GfxBuffer bvh_compact_size_readback_buffer_ = {};
+        std::vector<GfxGeometry> geometries_;
+    };
+    GfxArray<BottomLevelAccelerationStructure> bottom_level_acceleration_structures_;
+    GfxHandles bottom_level_acceleration_structure_handles_;
+    std::vector<std::vector<D3D12_RAYTRACING_GEOMETRY_DESC>> batch_descs_;
```
tengel-amd, line 496: shall we use vector or GfxArray?

gfx.cpp
```diff
@@ -444,64 +444,69 @@ class GfxInternal
...
+    std::vector<D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_INPUTS> batch_build_inputs_;
```
tengel-amd, line 497: `batch_descs_` and `batch_build_inputs_` are only temporarily used for building, right? then we shouldn't have to store them here.

gfx.cpp
```diff
@@ -2409,639 +2422,860 @@ class GfxInternal
...
+                instance_desc.Transform[j][3] = gfx_instance.transform_[4 * j + 3];
+            instance_desc.InstanceID = gfx_instance.instance_id_;
+            instance_desc.InstanceMask = gfx_instance.instance_mask_;
+            instance_desc.InstanceContributionToHitGroupIndex = gfx_instance.instance_contribution_to_hit_group_index_;
+            instance_desc.AccelerationStructure = gfx_buffer.resource_->GetGPUVirtualAddress() + gfx_buffer.data_offset_;
+            instance_descs[instance_desc_count++] = instance_desc;
+        }
+        if(instance_desc_count == 0)
+        {
+            destroyBuffer(gfx_acceleration_structure.bvh_buffer_);
```
tengel-amd, line 3105: empty scene should have a valid tlas

tengel-amd: question is who should handle this

gfx.cpp
```diff
@@ -2409,639 +2422,860 @@ class GfxInternal
...
+            auto const &instance = gfx_acceleration_structure.instances_[i];
+            if(!top_level_acceleration_structure_instance_handles_.has_handle(instance.handle))
+                continue;   // invalid tlas instance object
+            auto const &gfx_instance = top_level_acceleration_structure_instances_[instance];
+            GfxBuffer const& buffer = getTopLevelAccelerationStructureInstanceBuffer(gfx_instance);
+            if(!buffer_handles_.has_handle(buffer.handle))
+                continue;   // no valid BVH memory, probably wasn't built
+            D3D12_RAYTRACING_INSTANCE_DESC instance_desc = {};
+            Buffer const &gfx_buffer = buffers_[buffer];
+            for(uint32_t row = 0; row < 3; ++row)
```
tengel-amd: force the user to provide the right data so this is just a memcpy

gfx.cpp
```diff
@@ -2409,639 +2422,860 @@ class GfxInternal
...
+        if(!top_level_acceleration_structure_handles_.has_handle(tlas.handle))
+            return GFX_SET_ERROR(kGfxResult_InvalidOperation, "Cannot update an invalid top level acceleration structure object");
+        void* data = nullptr;
+        TopLevelAccelerationStructure &gfx_acceleration_structure = top_level_acceleration_structures_[tlas];
+        D3D12_GPU_VIRTUAL_ADDRESS const gpu_addr = allocateConstantMemory(gfx_acceleration_structure.instances_.size() * sizeof(D3D12_RAYTRACING_INSTANCE_DESC), data);
+        D3D12_RAYTRACING_INSTANCE_DESC *instance_descs = (D3D12_RAYTRACING_INSTANCE_DESC *)data;
+        uint32_t instance_desc_count = 0;
+        if(gpu_addr == 0)
+            return GFX_SET_ERROR(kGfxResult_OutOfMemory, "Unable to allocate for updating acceleration structure object with %u raytracing primitives", (uint32_t)gfx_acceleration_structure.instances_.size());
+        for(size_t i = 0; i < gfx_acceleration_structure.instances_.size(); ++i)
```
tengel-amd, line 3085: maybe not right now but ideally this kind of update would be done sparsely. There can be hundreds of thousands of elements here. There are only one or two field that can acually be updated, so you know what changed e.g., when a transform was updated, push_back the instance index to a vector, and sort to get the unique ones. Then you fill and upload a buffer of only updated instances and scatter them to the full buffer on GPU via compute shader. `fullInstanceBuffer[updatedInstances[did]] = updatedData[did]`

gfx.cpp
```diff
@@ -2409,639 +2422,860 @@ class GfxInternal
...
+            return GFX_SET_ERROR(kGfxResult_InvalidOperation, "Cannot destroy invalid top level acceleration structure object");
+        TopLevelAccelerationStructure const &gfx_tlas = top_level_acceleration_structures_[tlas];
+        collect(gfx_tlas); // release resources
+        top_level_acceleration_structures_.erase(tlas); // destroy tlas
+        top_level_acceleration_structure_handles_.free_handle(tlas.handle);
         return kGfxResult_NoError;
     }
 
-    GfxResult setRaytracingPrimitiveInstanceContributionToHitGroupIndex(GfxRaytracingPrimitive const &raytracing_primitive, uint32_t instance_contribution_to_hit_group_index)
+    GfxResult topLevelAccelerationStructureAddInstance(GfxTopLevelAccelerationStructure const &tlas, GfxTopLevelAccelerationStructureInstance const &instance)
```
tengel-amd, line 3020: since you're writing this new, you should already take an array as inputs for add/remove/updates e.g.,
`topLevelAccelerationStructureAddInstances(GfxTopLevelAccelerationStructure const &tlas, GfxTopLevelAccelerationStructureInstance const *instances, uint count)`. This is especially useful on the transform update hotpath `topLevelAccelerationStructureInstanceSetTransform`

gfx.cpp
```diff
@@ -2409,639 +2422,860 @@ class GfxInternal
...
+            gfx_acceleration_structure.bvh_buffer_ = createBuffer(bvh_data_size, nullptr, kGfxCpuAccess_None, D3D12_RESOURCE_STATE_RAYTRACING_ACCELERATION_STRUCTURE);
+            if(!gfx_acceleration_structure.bvh_buffer_)
+                return GFX_SET_ERROR(kGfxResult_OutOfMemory, "Unable to create acceleration structure buffer");
+        }
+        gfx_acceleration_structure.bvh_data_size_ = (uint64_t)tlas_info.ResultDataMaxSizeInBytes;
+        GFX_ASSERT(buffer_handles_.has_handle(gfx_acceleration_structure.bvh_buffer_.handle));
+        GFX_ASSERT(buffer_handles_.has_handle(raytracing_scratch_buffer_.handle));
+        Buffer &gfx_buffer = buffers_[gfx_acceleration_structure.bvh_buffer_];
+        Buffer &gfx_scratch_buffer = buffers_[raytracing_scratch_buffer_];
+        // TODO:
```
tengel-amd: TODO

gfx.cpp
```diff
@@ -2409,639 +2422,860 @@ class GfxInternal
...
+        //SetObjectName(gfx_buffer, acceleration_structure.name);
+        if(transitionResource(gfx_scratch_buffer, D3D12_RESOURCE_STATE_UNORDERED_ACCESS))
+            submitPipelineBarriers();   // ensure scratch is not in use
+        D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_DESC build_desc = {};
+        build_desc.DestAccelerationStructureData = gfx_buffer.resource_->GetGPUVirtualAddress() + gfx_buffer.data_offset_;
+        build_desc.Inputs = tlas_inputs;
+        if((tlas_inputs.Flags & D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PERFORM_UPDATE) != 0)
+            build_desc.SourceAccelerationStructureData = gfx_buffer.resource_->GetGPUVirtualAddress() + gfx_buffer.data_offset_;
+        build_desc.ScratchAccelerationStructureData = gfx_scratch_buffer.resource_->GetGPUVirtualAddress() + gfx_scratch_buffer.data_offset_;
+        GFX_ASSERT(dxr_command_list_ != nullptr);   // should never happen
```
tengel-amd, line 3143: no need for assert if it'll crash next line

gfx.cpp
```diff
@@ -2409,639 +2422,860 @@ class GfxInternal
...
+                    gfx_blas.bvh_compact_size_buffer_ = createBuffer(compact_size, nullptr, kGfxCpuAccess_None);
+                    gfx_blas.bvh_compact_size_readback_buffer_ = createBuffer(compact_size, nullptr, kGfxCpuAccess_Read);
+                    // We could pass `D3D12_RESOURCE_STATE_UNORDERED_ACCESS` to `createBuffer` directly,
+                    // but then DX12 spams warnings about "Ignoring InitialState D3D12_RESOURCE_STATE_UNORDERED_ACCESS".
+                    // So, in order to suppress them, we do this.
+                    Buffer &buffer = buffers_[gfx_blas.bvh_compact_size_buffer_];
+                    transitionResource(buffer, D3D12_RESOURCE_STATE_UNORDERED_ACCESS, kTransitionType_Implicit);
+                }
+            }
+            else if(gfx_blas.bvh_compact_size_buffer_)
```
tengel-amd, line 2837: isn't this an error (not allowing compaction and having `bvh_compact_size_buffer_`)?

gfx.cpp
```diff
@@ -5035,6 +5269,7 @@ class GfxInternal
         command_list_->Reset(command_allocators_[fence_index_], nullptr);
         resetState();   // re-install state
         decayResourceState();
+        finish_calls_counter_++;
```
tengel-amd: why is this done here?

gfx.cpp
```diff
@@ -2409,639 +2422,860 @@ class GfxInternal
...
+            return kGfxResult_NoError;
+        if(!bottom_level_acceleration_structure_handles_.has_handle(blas.handle))
+            return GFX_SET_ERROR(kGfxResult_InvalidOperation, "Cannot compact invalid bottom level acceleration structure object");
+        BottomLevelAccelerationStructure &gfx_blas = bottom_level_acceleration_structures_[blas];
+        bool const allow_compaction = (gfx_blas.build_flags_ & kGfxBuildBottomLevelASFlag_Compact) != 0;
+        if(!allow_compaction)
+            return GFX_SET_ERROR(kGfxResult_InvalidOperation, "Compaction is not allowed for this bottom level acceleration structure object");
+        if(gfx_blas.finish_calls_state_ >= finish_calls_counter_)
+            return GFX_SET_ERROR(kGfxResult_InvalidOperation, "Post-build information is not available yet. Please, call `gfxFinish()' and try again");
+        Buffer &dst = buffers_[gfx_blas.bvh_compact_size_readback_buffer_];
```
tengel-amd, line 2671: how do we guarantee that the readback data was written?

gfx.cpp
VladimirMakeew: They stored in `CapsaicinInternal` to preserve allocated memory and remove unnecessary reallocations. See: https://github.com/gboisse/gfx/pull/218#discussion_r3580587257

gfx.cpp
VladimirMakeew: It's the same old behavior: https://github.com/gboisse/gfx/blob/master/gfx.cpp#L2490

gfx.cpp
VladimirMakeew: Thats what `finish_calls_counter_` is for. To make sure user called `gfxFinish` between build and compact.
Fence will be a better solution, tho

tengel-amd: This should be done async, you know it's going to be finished in `max_frames_in_flight_`, then you can read and start the compaction. No need to block for this.

gfx.cpp
VladimirMakeew: `GfxArray` is a sparse array container, vector is better for this case, I think.

gfx.cpp
VladimirMakeew: No, in case if blas geometry or build flags were changed. If no compaction takes place, compaction buffers should not waste resources.

gfx.h
VladimirMakeew: Done. Batch build and update functions were also renamed

gfx.cpp
VladimirMakeew: Done

gfx.cpp
VladimirMakeew: Fixed

gfx.cpp
VladimirMakeew: Fixed, now checking its existence along with `dxr_device` at the start of the function

gfx.cpp
tengel-amd: I would not allow changing that in the first place 😬

gfx.cpp
VladimirMakeew: Fixed, did the same way as in #227 since my compaction logic is based on #213

gfx.cpp
```diff
@@ -2409,639 +2423,711 @@ class GfxInternal
...
+            GfxBottomLevelAccelerationStructure const &blas = blases[i];
+            BottomLevelAccelerationStructure &gfx_blas = bottom_level_acceleration_structures_[blas];
+            if(!update && flags != nullptr)
+                gfx_blas.build_flags_ = flags[i];
+            std::vector<D3D12_RAYTRACING_GEOMETRY_DESC> blas_descs;
+            blas_descs.reserve(gfx_blas.geometries_.size());
+            for(size_t j = 0; j < gfx_blas.geometries_.size(); ++j)
+            {
+                GfxGeometry const &geometry = gfx_blas.geometries_[j];
+                if(!geometry_handles_.has_handle(geometry.handle))
```
tengel-amd: could move with the other checks above

VladimirMakeew: Done

gfx.cpp
VladimirMakeew: Discussed in Slack. From the gfx library perspective user can change geometries of a built or compacted blas.
I've added `BottomLevelAccelerationStructure::state_` member variable to keep track of `need build` -> `built` -> `compacted` blas states. Adding or removing geometries is possible, but requires user to perform a fresh build afterwards.

gfx.cpp
```diff
@@ -2409,639 +2432,729 @@ class GfxInternal
...
-        gfx_raytracing_primitive.procedural_.acceleration_structure_ = acceleration_structure;
-        gfx_raytracing_primitive.instance_id_ = raytracing_primitive.getIndex();
-        gfx_raytracing_primitive.type_ = RaytracingPrimitive::kType_Procedural;
-        gfx_acceleration_structure.needs_rebuild_ = true;
-        return raytracing_primitive;
+        return &top_level_acceleration_structure_instances_[instance].blas_;
     }
 
-    GfxResult destroyRaytracingPrimitive(GfxRaytracingPrimitive const &raytracing_primitive)
+    GfxResult topLevelAccelerationStructureInstanceSetBottomLevelAccelerationStructure(GfxTopLevelAccelerationStructureInstance const &instance, GfxBottomLevelAccelerationStructure const &blas)
```
tengel-amd, line 2962: it's a bit of boilerplate these functions, is it really needed or could we set directly a struct with all fields? remember that this may be called many times in a frame and we should avoid updating field by field in such cases.

gfx.cpp
tengel-amd: assertion is still here though

gfx.cpp
tengel-amd: I really wonder if we should even have this add/remove functionalities. I'd even go for a simply `topLevelAccelerationStructureSetInstances` since there is nothing we need to keep track of on instance side (only BLAS itself holds a resource), we can directly send this off to the TLAS instance buffer and we're done.

VladimirMakeew: One of the requirements for new api is the ability to switch BLAS's

gfx.cpp
VladimirMakeew: Removed

gfx.cpp
VladimirMakeew: Removed, `dxr_device` is also checked at the top of `topLevelAccelerationStructureBuild`

gfx.cpp
VladimirMakeew: This won't help that much, these are just operations on `vector` inside TLAS object. Each instance should be checked for validity before performing actual add/remove.
Biggest speed up was achieved by reducing barriers count and batching BLAS builds. Next potential improvement, as you suggested, could be using shaders to populate build data.

gfx.cpp
tengel-amd: you can still do that, there is no tracking needed. Caller is still responsible for keeping it alive. 
Anyway let's put a pin on this, but I think if we update this on GPU this API would go away anyways.

tengel-amd (approved):

gboisse:
> Fixed, did the same way as in https://github.com/gboisse/gfx/pull/227 since my compaction logic is based on https://github.com/gboisse/gfx/pull/213

Does this mean this PR supersedes https://github.com/gboisse/gfx/pull/213? or should we still merge both?

VladimirMakeew:
@gboisse this PR supersedes #213

gfx.h
```diff
@@ -194,63 +194,88 @@ GfxSamplerState gfxCreateSamplerState(GfxContext context, D3D12_FILTER filter, D
...
+//!
+//! Top level acceleration structure instance
+//!
+
+class GfxTopLevelAccelerationStructureInstance { GFX_INTERNAL_HANDLE(GfxTopLevelAccelerationStructureInstance); public: };
+
+GfxTopLevelAccelerationStructureInstance gfxCreateTopLevelAccelerationStructureInstance(GfxContext context);
+GfxResult gfxDestroyTopLevelAccelerationStructureInstance(GfxContext context, GfxTopLevelAccelerationStructureInstance instance);
+
+GfxBottomLevelAccelerationStructure const *gfxTopLevelAccelerationStructureInstanceGetBottomLevelAccelerationStructure(GfxContext context, GfxTopLevelAccelerationStructureInstance instance);
```
gboisse: Does this return a list or a single element?
If a list, the function should this be spelt in plural, if not, there's no point in returning a pointer.

gboisse:
Enjoying the diff tally on that one 🙂
image

gboisse:
Changes are looking good but break the examples, so let's fix this then we can merge:
image

To repro, you can clone `gfx` on its own, enable the various features (i.e., `GFX_ENABLE_GUI` and `GFX_ENABLE_SCENE`), then try to build and run the "rtao" and "pbr" examples.

maoliver-amd:
> Changes are looking good but break the examples, so let's fix this then we can merge: 
> To repro, you can clone `gfx` on its own, enable the various features (i.e., `GFX_ENABLE_GUI` and `GFX_ENABLE_SCENE`), then try to build and run the "rtao" and "pbr" examples.

Alternatively you can just add `"GFX_BUILD_EXAMPLES": true` to your CMakePresets.json in the "cacheVariables" section

gfx.h
```diff
@@ -497,11 +522,11 @@ GfxResult gfxResetCommandListState(GfxContext context); // call this function be
 
 GfxBuffer gfxCreateBuffer(GfxContext context, ID3D12Resource *resource, D3D12_RESOURCE_STATES resource_state = D3D12_RESOURCE_STATE_PIXEL_SHADER_RESOURCE);
 GfxTexture gfxCreateTexture(GfxContext context, ID3D12Resource *resource, D3D12_RESOURCE_STATES resource_state = D3D12_RESOURCE_STATE_PIXEL_SHADER_RESOURCE);
-GfxAccelerationStructure gfxCreateAccelerationStructure(GfxContext context, ID3D12Resource *resource, uint64_t byte_offset = 0);    // resource must be in D3D12_RESOURCE_STATE_RAYTRACING_ACCELERATION_STRUCTURE state
+GfxTopLevelAccelerationStructure gfxCreateTopLevelAccelerationStructure(GfxContext context, ID3D12Resource *resource, uint64_t byte_offset = 0);    // resource must be in D3D12_RESOURCE_STATE_RAYTRACING_ACCELERATION_STRUCTURE state
 
 ID3D12Resource *gfxBufferGetResource(GfxContext context, GfxBuffer buffer);
 ID3D12Resource *gfxTextureGetResource(GfxContext context, GfxTexture texture);
-ID3D12Resource *gfxAccelerationStructureGetResource(GfxContext context, GfxAccelerationStructure acceleration_structure);
+ID3D12Resource *gfxAccelerationStructureGetResource(GfxContext context, GfxTopLevelAccelerationStructure tlas);
```
gboisse: This should probably be renamed to `gfxTopLevelAccelerationStructureGetResource()` for better consistency.

gfx.cpp
```diff
@@ -2409,639 +2435,729 @@ class GfxInternal
...
-    }
-
-    GfxResult buildRaytracingPrimitiveProcedural(GfxRaytracingPrimitive const &raytracing_primitive, GfxBuffer const &aabb_buffer, uint32_t aabb_stride, GfxBuildRaytracingPrimitiveFlags build_flags)
+        if(!instance)
+            return kGfxResult_NoError;
+        if(!top_level_acceleration_structure_instance_handles_.has_handle(instance.handle))
+            return GFX_SET_ERROR(kGfxResult_InvalidOperation, "Cannot destroy invalid top level acceleration structure instance object");
+        TopLevelAccelerationStructureInstance const &gfx_instance = top_level_acceleration_structure_instances_[instance];
+        collect(gfx_instance); // release resources
+        top_level_acceleration_structure_instances_.erase(instance); // destroy blas
```
gboisse: Comment should read `// destroy tlas` or `// destroy tlas instance`?

gfx.h
gboisse: ha, turns out it's properly renamed in the `.cpp` but not in the header; need to fix in header.

gfx.cpp
```diff
@@ -2409,639 +2435,729 @@ class GfxInternal
...
+    GfxGeometry createGeometryTriangles(GfxBuffer const &vertex_buffer, uint32_t vertex_stride)
+    {
+        GfxGeometry geometry = {};
+        geometry.type = GfxGeometry::kType_Triangles;
+        geometry.handle = geometry_handles_.allocate_handle();
+        Geometry &gfx_geometry = geometries_.insert(geometry);
+        gfx_geometry.type_ = Geometry::kType_Triangles;
+        gfx_geometry.data_.triangles_.index_buffer_ = {};
+        gfx_geometry.data_.triangles_.index_stride_ = 0;
+        gfx_geometry.data_.triangles_.vertex_buffer_ = vertex_buffer;
```
gboisse, line 2446: Previous implementation was creating a buffer range over the vertex/index buffers, e.g.:
```
gfx_geometry.data_.triangles_.vertex_buffer_ = createBufferRange(vertex_buffer, 0, vertex_buffer.size);
```

The point is that creating a buffer range increments the reference count of the vertex/index buffer so it can be safely deleted on the user side.

I'm fine with the behaviour change (as it wasn't clear or obvious anyway in the previous version), but we need to make sure to handle correctly the error case where a geometry is created with a vertex buffer, the buffer is deleted, then the user tries and build the BLAS.

gfx.h
```diff
@@ -194,63 +194,88 @@ GfxSamplerState gfxCreateSamplerState(GfxContext context, D3D12_FILTER filter, D
...
+uint64_t gfxBottomLevelAccelerationStructureGetDataSize(GfxContext context, GfxBottomLevelAccelerationStructure blas); // in bytes
+
+GfxResult gfxBuildBottomLevelAccelerationStructures(GfxContext context, GfxBottomLevelAccelerationStructure const *blases, GfxBuildBottomLevelASFlags const *flags, uint32_t blas_count);
+GfxResult gfxUpdateBottomLevelAccelerationStructures(GfxContext context, GfxBottomLevelAccelerationStructure const *blases, uint32_t blas_count);
+
+//!
+//! Top level acceleration structure instance
+//!
+
+class GfxTopLevelAccelerationStructureInstance { GFX_INTERNAL_HANDLE(GfxTopLevelAccelerationStructureInstance); public: };
```
gboisse, line 250: Any reason for this not to be using `GFX_INTERNAL_NAMED_HANDLE`? is there no underlying D3D12 resource corresponding with this object?

gboisse: Seems there isn't underlying resources indeed 🙂
https://github.com/gboisse/gfx/blob/4878fdbb9f2faf12f0c68f0974e46e93b9e2ce43/gfx.cpp#L6311-L6314

That said, it could still be useful for debugging to have a name handy.

As in, I could imagine setting the glTF instance name to the handle, which could simplify step-through debugging in some cases?

Just thinking out loud, not saying it needs changing, happy to hear your thoughts on this.

gfx.cpp
```diff
@@ -444,64 +445,81 @@ class GfxInternal
...
+                uint32_t vertex_stride_ = 0;
+                GfxBuffer vertex_buffer_ = {};
+            } triangles_;
+            struct
+            {
+                uint32_t procedural_stride_ = 0;
+                GfxBuffer procedural_buffer_ = {};
+            } procedural_;
+        } data_ = {};
+        bool opaque = false;
```
gboisse: Missing trailing underscore, i.e., should be named `opaque_`.

VladimirMakeew: Fixed

gfx.cpp
VladimirMakeew: Fixed the comment, we are destroying tlas instance here

gfx.h
VladimirMakeew: Oops! Fixed

gfx.h
VladimirMakeew: It returns a single blas from tlas instance. Fixed return type

gfx.h
VladimirMakeew: I added names support to BLAS and TLAS only and did that mostly for GPU debug purposes.
As BVH complexity grows with introduction of compaction, multiple geometries, blas swap (and OMM + CLAS in the future) I thought it would be useful to find things quickly in PIX or RenderDoc.
Later we can introduce something like `GFX_INTERNAL_DEBUG_NAME_HANDLE` to remove debug names from release builds at compile time to save some memory and cpu cycles.

Debugging with PIX. It also has BVH 3d model view which also helps:
bvh_debug_names

VladimirMakeew:
@gboisse I've fixed pbr and rtao examples compilation and BVH build logic.
PBR looks the same as master branch while RTAO is different.
| Master | This PR |
|---------|---------|
| rtao_master | rtao_branch |

Render with debug materials to see geometry:
rtao_branch_debug_colors
It behaves as if AO radius is always set to 0. I tried hardcoding non zero values in the shader but results looked the same.
Scene hierarchy results in a single BLAS with about 2k geometries. I tried changing the build to produce 2k BLAS's each with a single geometry - no visible changes.
Don't know why I'm getting such result, need some time to debug.

gboisse:
> @gboisse I've fixed pbr and rtao examples compilation and BVH build logic. PBR looks the same as master branch while RTAO is different.
> Master 	This PR
> rtao_master 	rtao_branch
> 
> Render with debug materials to see geometry: rtao_branch_debug_colors It behaves as if AO radius is always set to 0. I tried hardcoding non zero values in the shader but results looked the same. Scene hierarchy results in a single BLAS with about 2k geometries. I tried changing the build to produce 2k BLAS's each with a single geometry - no visible changes. Don't know why I'm getting such result, need some time to debug.

I wonder if it could be caused by some invalid combination of flags in the ray query object?
https://github.com/gboisse/gfx/blob/973ae71d81cd393cda48f6525ae09ca5d9438cc1/examples/01-rtao/rtao.frag#L119-L121

Although it is strange that it would work on `master` but not on the branch.

VladimirMakeew:
@gboisse RTAO example is now fixed.
TLAS instances should have identity transform by default or else ray tracing doesn't work as expected.
It was also noticeable in Radeon Raytracing Analyzer, it showed proper BLAS's with geometry while whole TLAS 3d view was empty.
After the fix:
rtao_branch_fixed

gboisse:
> @gboisse RTAO example is now fixed. TLAS instances should have identity transform by default or else ray tracing doesn't work as expected. It was also noticeable in Radeon Raytracing Analyzer, it showed proper BLAS's with geometry while whole TLAS 3d view was empty. After the fix: rtao_branch_fixed

Nice!

gfx.cpp
VladimirMakeew: I've added validity checks for geometry buffers right before BLAS build: https://github.com/gboisse/gfx/commit/72d350aab89a874090d3f7f0e16e38273730d9c2
Now cases as such are handled correctly:
```c++
GfxBottomLevelAccelerationStructure blas = // ...

GfxGeometry geometry = gfxCreateGeometryTriangles(gfx, index, vertex, sizeof(Vertex));
gfxBottomLevelAccelerationStructureAddGeometry(gfx, blas, geometry);
// ...
gfxDestroyBuffer(gfx, index);
gfxDestroyBuffer(gfx, vertex);
// Build will fail and return kGfxResult_InvalidParameter
gfxBuildBottomLevelAccelerationStructures(gfx, &blas, &build_flags, 1u);
```

gboisse (approved):

PR #222, opened 2026-07-21
Update BLAS with multiple geometries

VladimirMakeew:
Add new api function `gfxRaytracingPrimitiveUpdateBatch` which allows to update multiple `GfxRaytracingPrimitive`-s that were previously built into a single BLAS.

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -2741,7 +2741,7 @@ public:
         return buildRaytracingPrimitive(raytracing_primitive, gfx_raytracing_primitive, false);
     }
 
-    GfxResult buildRaytracingPrimitiveBatch(GfxRaytracingPrimitiveBatchElement const* batch, size_t batch_size)
+    GfxResult buildRaytracingPrimitiveBatch(GfxRaytracingPrimitiveBatchElement const* batch, size_t batch_size, bool update)
     {
         if(dxr_device_ == nullptr)
             return kGfxResult_InvalidOperation; // avoid spamming console output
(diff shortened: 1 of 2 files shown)


gfx.cpp
tengel-amd, line 2778: I wonder if we can assume that we have the same buffer for all elements in the batch? then we could do something like done below but only once for the whole batch and no need to create new vertex/index buffer for every element in the batch.
`gfx_index_buffer->resource_->GetGPUVirtualAddress() + gfx_index_buffer->data_offset_; `

gfx.cpp
```diff
@@ -2830,6 +2830,8 @@ class GfxInternal
         blas_inputs.Flags = D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_UPDATE;
```
tengel-amd, line 2830: `D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_UPDATE` should not be needed for static geometry for example. Or are we assuming these batches are only for dynamic geometry?

gfx.cpp
tengel-amd: ` uint64_t const scratch_data_size = update ? blas_info.UpdateScratchDataSizeInBytes : blas_info.ScratchDataSizeInBytes;` adjust the scratch accordingly?

tengel-amd: Also this should be split. First we check how much scratch memory we need, then we `allocateRaytracingScratch` once, to avoid possibly having multiple allocations on a batch call.

gfx.cpp
VladimirMakeew: This is the same logic we currently have in `gfxRaytracingPrimitiveUpdate`. There is a PR with proper [build flags](https://github.com/gboisse/gfx/pull/213) support. And [my other PR](https://github.com/gboisse/gfx/pull/219) will also have them along with fast builds and different blas packing api.

gfx.cpp
VladimirMakeew: Since this is an library api function we shouldn't assume that. We can add another api function for the specific case you mentioned.

tengel-amd: well the user can always choose not to use the batch version

tengel-amd (approved):

gfx.cpp
maoliver-amd: Were actually looking to move to have separate buffers for each primitive, however, we could still batch geometries that shares single render instance into a single buffer. Is there much to gain from forcing all to be in the same buffer? We could add that as a function constraint

maoliver-amd: or just wait for the new API?

tengel-amd: It doesn't have to be done now, it's just something that came to mind. Surely allocating and deallocating these buffers for every blas every frame doesn't scale well (even if memory itself may already be allocated), and the barriers could then be simplified if we could make such assumptions.

VladimirMakeew: Fixed

PR #224, opened 2026-08-06
Make `gfx` thread safe (except for ImGui integration)

gboisse:
This should make `gfx` thread safe, as in, multiple threads can each have their own `GfxContext` and operate without synchronization.

It does not allow to use a single `GfxContext` concurrently onto multiple threads however.

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -9896,12 +9896,12 @@ private:
                     {
                         fwrite(dxc_bytecode->GetBufferPointer(), dxc_bytecode->GetBufferSize(), 1, fd);
                         fclose(fd); // write out bytecode for shader caching
-                    }
-                    fd = _wfopen(shader_key_reflection.c_str(), L"wb");
-                    if(fd)
-                    {
-                        fwrite(dxc_reflection->GetBufferPointer(), dxc_reflection->GetBufferSize(), 1, fd);
-                        fclose(fd); // write out reflection for shader caching
+                        fd = _wfopen(shader_key_reflection.c_str(), L"wb");
+                        if(fd)
+                        {
+                            fwrite(dxc_reflection->GetBufferPointer(), dxc_reflection->GetBufferSize(), 1, fd);
+                            fclose(fd); // write out reflection for shader caching
+                        }
                     }
                     // Delete older cached files
                     std::filesystem::directory_iterator const end;
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -367,7 +367,7 @@ class GfxInternal
             } raster_state_;
         };
 
-        DrawState() : reference_count_(0) {}
+        DrawState() : reference_count_(1) {}    // retain
         DrawState(DrawState &&other) : draw_state_(other.draw_state_), reference_count_(other.reference_count_) { other.reference_count_ = 0; }
         ~DrawState() { GFX_ASSERT(reference_count_ == 0); }
 
@@ -386,8 +386,6 @@ class GfxInternal
         Data draw_state_;
         uint32_t reference_count_;
     };
-    static GfxArray<DrawState> draw_states_;
-    static GfxHandles draw_state_handles_;
 
     struct Object
     {
@@ -3196,8 +3194,7 @@ public:
             return mesh_kernel;
         }
         GFX_ASSERT(define_count == 0 || defines != nullptr);
-        uint32_t const draw_state_index = static_cast<uint32_t>(draw_state.handle & 0xFFFFFFFFull);
-        DrawState const *gfx_draw_state = draw_states_.at(draw_state_index);
+        DrawState const *gfx_draw_state = reinterpret_cast<DrawState *>(draw_state.handle);
         if(!gfx_draw_state)
         {
             GFX_PRINT_ERROR(kGfxResult_InvalidOperation, "Cannot create a mesh kernel using an invalid draw state object");
@@ -3267,8 +3264,7 @@ public:
             return graphics_kernel;
         }
         GFX_ASSERT(define_count == 0 || defines != nullptr);
-        uint32_t const draw_state_index = static_cast<uint32_t>(draw_state.handle & 0xFFFFFFFFull);
-        DrawState const *gfx_draw_state = draw_states_.at(draw_state_index);
+        DrawState const *gfx_draw_state = reinterpret_cast<DrawState *>(draw_state.handle);
         if(!gfx_draw_state)
         {
             GFX_PRINT_ERROR(kGfxResult_InvalidOperation, "Cannot create a graphics kernel using an invalid draw state object");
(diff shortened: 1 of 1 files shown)


gfx.cpp
```diff
@@ -367,7 +367,7 @@ class GfxInternal
             } raster_state_;
         };
 
-        DrawState() : reference_count_(0) {}
+        DrawState() : reference_count_(1) {}    // retain
```
gboisse, line 370: Retains automatically, rather than having to set `reference_count_` to `1` in `DispenseDrawState()`.

gfx.cpp
```diff
@@ -386,8 +386,6 @@ class GfxInternal
         Data draw_state_;
         uint32_t reference_count_;
     };
-    static GfxArray<DrawState> draw_states_;
-    static GfxHandles draw_state_handles_;
```
gboisse, line 390: This was the major source of thread "unsafety", removed completely in favor of malloc and free of the internal `DrawState` objects.

gfx.cpp
```diff
@@ -3196,8 +3194,7 @@ class GfxInternal
             return mesh_kernel;
         }
         GFX_ASSERT(define_count == 0 || defines != nullptr);
-        uint32_t const draw_state_index = static_cast<uint32_t>(draw_state.handle & 0xFFFFFFFFull);
-        DrawState const *gfx_draw_state = draw_states_.at(draw_state_index);
+        DrawState const *gfx_draw_state = reinterpret_cast<DrawState *>(draw_state.handle);
```
gboisse, line 3197: `handle` is now just a pointer to the internal (heap-allocated) `DrawState` object.

gfx.cpp
```diff
@@ -5576,38 +5572,30 @@ class GfxInternal
 
     static void DispenseDrawState(GfxDrawState &draw_state)
     {
-        draw_state.handle = draw_state_handles_.allocate_handle();
-        uint32_t const draw_state_index = static_cast<uint32_t>(draw_state.handle & 0xFFFFFFFFull);
-        GFX_ASSERT(!draw_states_.has(draw_state_index));    // should never happen
-        draw_states_.insert(draw_state_index).reference_count_ = 1;
+        draw_state.handle = reinterpret_cast<uint64_t>(new DrawState());
+        GFX_ASSERT(!!draw_state.handle);    // out of memory
```
gboisse, line 5576: malloc...

gfx.cpp
```diff
@@ -5576,38 +5572,30 @@ class GfxInternal
...
-        DrawState *gfx_draw_state = draw_states_.at(draw_state_index);  // look up draw state
+        DrawState *gfx_draw_state = reinterpret_cast<DrawState *>(draw_state.handle);
         GFX_ASSERT(gfx_draw_state != nullptr); if(!gfx_draw_state) return;
         GFX_ASSERT(gfx_draw_state->reference_count_ > 0);
         if(--gfx_draw_state->reference_count_ == 0)
-        {
-            draw_states_.erase(draw_state_index);
-            draw_state_handles_.free_handle(draw_state.handle);
-        }
+            delete gfx_draw_state;
```
gboisse, line 5593: ... and free.

gfx.cpp
```diff
@@ -9915,12 +9896,12 @@ class GfxInternal
...
-                    if(fd)
-                    {
-                        fwrite(dxc_reflection->GetBufferPointer(), dxc_reflection->GetBufferSize(), 1, fd);
-                        fclose(fd); // write out reflection for shader caching
+                        fd = _wfopen(shader_key_reflection.c_str(), L"wb");
+                        if(fd)
+                        {
+                            fwrite(dxc_reflection->GetBufferPointer(), dxc_reflection->GetBufferSize(), 1, fd);
+                            fclose(fd); // write out reflection for shader caching
+                        }
```
gboisse, line 9904: Slightly updated this to **not** write the reflection data if the DXIL file couldn't be written.

Other than that, it's looking like the rest of the logic should already be thread safe?

maoliver-amd:
Nice, Looks like this should solve the drawstate issue.

There are still a couple of other static variables that can cause race conditions:
`static bool created_shader_cache_directory;` -> atomic bool with compare_exchange or mutex?
`static bool created_shader_pdb_directory;`
`static bool warned;` -> remove and just return error?
`static GfxMessageCallback userMessageCallback;` -> was never made per context like originally talked about

gboisse:
`static bool created_shader_cache_directory;` -> my reasoning here was that the subsequent kernel calls are thread safe and should simply return errors such as "already exist", which we treat as "no error".
Might be worth being more explicit potentially, but maybe what's in place is enough to run correctly?

[EDIT] one thing I'm unsure of looking at the code again, is the eviction logic relying on `std::filesystem` (could that throw uncaught exceptions in some cases?).

`static bool warned;` -> I did notice that one, not really a big deal tho I would say. Worst case I am aware of is that multiple threads could race and user would see the message being printed out a few times rather than only once.

gboisse:
`static GfxMessageCallback userMessageCallback;` -> mmh I believe there was a discussion there, but it seems it unfortunately didn't happen on GitHub... But I recall we ultimately decided to keep the global version, will try and find the discussion back.

One more "proper" solution could be to have the ability to (thread-safely) register (and unregister) callbacks; then it wouldn't be the last to set it "wins".

maoliver-amd:
> `static bool created_shader_cache_directory;` -> my reasoning here was that the subsequent kernel calls are thread safe and should simply return errors such as "already exist", which we treat as "no error". Might be worth being more explicit potentially, but maybe what's in place is enough to run correctly?

True, the `_wmkdir` will just return `EEXIST` and everything will continue as normal

> [EDIT] one thing I'm unsure of looking at the code again, is the eviction logic relying on `std::filesystem` (could that throw uncaught exceptions in some cases?).

It is possible with the code in general that we could have multiple contexts trying to compile the same shader in parallel resulting in race conditions for reading/writing. For the filesystem case we can use the error_code variants which will silently just ignore errors and continue, which in this case is perfectly acceptable outcome.
changing code to:
```
// Delete older pdb files
std::filesystem::directory_iterator const end;
std::vector<std::filesystem::path>        pdb_files;
std::error_code                           ec, ec2;
for(std::filesystem::directory_iterator iter{std::filesystem::path(shader_pdb_dir + shader_key_dir)}; iter != end && !ec2; iter.increment(ec2))
    if(is_regular_file(*iter, ec))
        if(iter->path().extension() == ".pdb")
            pdb_files.emplace_back(*iter);
if(pdb_files.size() > max_cached_files)
{
    std::sort(pdb_files.begin(), pdb_files.end(), [&ec](const std::filesystem::path &a, const std::filesystem::path &b) {
        return last_write_time(a, ec) > last_write_time(b, ec); });
    for(size_t i = max_cached_files; i < pdb_files.size(); ++i)
        std::filesystem::remove(pdb_files[i], ec);
}
```
and the same for the `bytecode_files` and `reflection_files`. `is_regular_file` will just return false on error which prevent the file being added to the list, `last_write_time` will return a bogus time value if the file no longer exists which will result in the file being sorted to the end of the list `filesystem::remove` will then just silently ignore these already deleted files.
This should now gracefully handle most race conditions. There is still a chance of OS implementation defined exceptions which will cause a terminate. The only way to fix that is with a lock

> `static bool warned;` -> I did notice that one, not really a big deal tho I would say. Worst case I am aware of is that multiple threads could race and user would see the message being printed out a few times rather than only once.

Yep fair enough

gfx.cpp
tengel-amd: (maybe I don't understand the codebase well enough) but would it be possible (or what would it take) to get rid of the heap allocation altogether?  Seems like it just stores 80 bytes and is copied anyway into the kernel upon kernel creation. I did a quick try here and didn't run into issues at least by moving `DrawState` to the header and renaming it to `GfxDrawState`, removing the ref counting and changing `gfxDrawStateSet*` methods to receive a reference type

gboisse: That'd be possible indeed; the idea was to keep the "opaque" nature of the structure (so it could be changed internally freely without breaking the user-level API for instance).

Ultimately my opinion here is that the draw state API is only used whenever the user is about to create a graphics and/or mesh kernel(s).
That is, invoking the shader compiler, reflection functionalities, file system, etc. So a heavy operation, where an additional malloc and free has little impact on the overall performance (i.e., this is something you'd only do upon initializing/terminating a module rather than per frame, where we ideally want no malloc/free happening at all).

tengel-amd: > That'd be possible indeed; the idea was to keep the "opaque" nature of the structure (so it could be changed internally freely without breaking the user-level API for instance).

that would be still the case since the `gfxDrawStateSet*` still work the same way and no interaction with internals is needed.

But yeah, it's not a hot path so performance is not a concern here. It would simplify this code a lot though, and also solve the threading/multiple context issue. Up to you if you rather keep it.

tengel-amd (approved):

PR #225, opened 2026-08-10
Fix ImGUI rendering with examples

maoliver-amd:
The current 01-rtao example experiences corrupted UI output. The cause is due to an incorrect texture being bound to the imgui color target. After the accumulate pass (gfxCommandBindColorTarget(gfx, 0, accum_buffer)), accum_buffer remains bound to color target slot 0. When gfxImGuiRender() calls installShaderState, it finds a non-empty texture in slot 0 and the imgui kernel's draw state declares DXGI_FORMAT_R8G8B8A8_UNORM_SRGB. This causes ImGui to render into accum_buffer (DXGI_FORMAT_R32G32B32A32_FLOAT) — a format mismatch with the PSO, producing the visual corruption.

The fix is to allow unbinding targets through the gfx API. the gfx imgui code can then unbind the color target when not in use.

### DIFF
--- a/examples/01-rtao/rtao.frag
+++ b/examples/01-rtao/rtao.frag
@@ -136,5 +136,5 @@ float4 Resolve(in float4 pos : SV_Position) : SV_Target
     float4 color_sample = AccumBuffer.Load(int3(pos.xy, 0));
     float3 color_value  = color_sample.xyz / max(color_sample.w - 1.0f, 1.0f);
 
-    return float4(color_value, 1.0f);
+    return float4(color_value, 0.0f);
 }
--- a/examples/02-pbr/pbr.frag
+++ b/examples/02-pbr/pbr.frag
@@ -162,7 +162,7 @@ Result main(in Params params)
 
     // Populate our multiple render targets (i.e., MRT)
     Result result;
-    result.color    = float4(color, 0.0f);
+    result.color    = float4(color, 1.0f);
     result.velocity = CalculateVelocity(params);
 
     return result;
--- a/examples/02-pbr/pbr.frag
+++ b/examples/02-pbr/pbr.frag
@@ -162,7 +162,7 @@ Result main(in Params params)
 
     // Populate our multiple render targets (i.e., MRT)
     Result result;
-    result.color    = float4(color, 1.0f);
+    result.color    = float4(color, 0.0f);
     result.velocity = CalculateVelocity(params);
 
     return result;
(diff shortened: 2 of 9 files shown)


examples/01-rtao/rtao.frag
```diff
@@ -136,5 +136,5 @@ float4 Resolve(in float4 pos : SV_Position) : SV_Target
     float4 color_sample = AccumBuffer.Load(int3(pos.xy, 0));
     float3 color_value  = color_sample.xyz / max(color_sample.w - 1.0f, 1.0f);
 
-    return float4(color_value, 1.0f);
+    return float4(color_value, 0.0f);
```
maoliver-amd, line 139: Fixes alpha blending of UI

examples/02-pbr/pbr.frag
```diff
@@ -162,7 +162,7 @@ Result main(in Params params)
 
     // Populate our multiple render targets (i.e., MRT)
     Result result;
-    result.color    = float4(color, 1.0f);
+    result.color    = float4(color, 0.0f);
```
maoliver-amd: 02-pbr doesnt render any UI despite still calling the imgui initialization and render functions each frame. In case these ever get used the output alpha was updated as needed

gfx_imgui.cpp
```diff
@@ -589,12 +600,12 @@ class GfxImGuiInternal
     static ImDrawCallback GetNormalAlphaCallback() { return gfx_DrawCallback_SetNormalAlpha; }
 };
 
-GfxResult gfxImGuiInitialize(GfxContext gfx, char const **font_filenames, uint32_t font_count, ImFontConfig const *font_configs, ImGuiConfigFlags flags)
+GfxResult gfxImGuiInitialize(GfxContext gfx, bool render_to_texture, char const **font_filenames, uint32_t font_count, ImFontConfig const *font_configs, ImGuiConfigFlags flags)
```
maoliver-amd: To ensure correct handling of imgui draw state, the api needed updating such that it knows ahead of time whether it will render to the back buffer directly or to an output texture

examples/02-pbr/pbr.frag
gboisse: ha, that's right; I've removed the ImGui code+include from the example.

gfx.cpp
```diff
@@ -3991,7 +3991,10 @@ class GfxInternal
         if(target_index >= kGfxConstant_MaxRenderTarget)
             return GFX_SET_ERROR(kGfxResult_InvalidParameter, "Cannot bind more than %u render targets", (uint32_t)kGfxConstant_MaxRenderTarget);
         if(!target_texture)
-            return GFX_SET_ERROR(kGfxResult_InvalidOperation, "Cannot draw to an invalid texture object");
+        {
+            bound_color_targets_[target_index] = {}; // unbind
+            return kGfxResult_NoError;
+        }
```
gboisse, line 3997: This is a good catch actually.

gboisse (approved):

PR #226, opened 2026-08-10
Add mutexes around filesystem operations

maoliver-amd:
This adds locking around filesystem ops when writing shader cache and pdbs to avoid conflicts with multiple threads

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -9948,7 +9948,7 @@ private:
                         for(size_t i = max_cached_files; i < bytecode_files.size(); ++i)
                             std::filesystem::remove(bytecode_files[i], ec);
                     }
-                    if (reflection_files.size() > max_cached_files)
+                    if(reflection_files.size() > max_cached_files)
                     {
                         std::sort(reflection_files.begin(), reflection_files.end(),
                             [&ec](std::filesystem::path const &a, std::filesystem::path const &b) {
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -9880,19 +9880,24 @@ private:
                 fclose(fd); // write out PDB for shader debugging
             }
             // Delete older pdb files
-            std::filesystem::directory_iterator const end;
-            std::vector<std::filesystem::path>        pdb_files;
-            std::error_code                           ec, ec2;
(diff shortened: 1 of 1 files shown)


gfx.cpp
```diff
@@ -9880,18 +9881,21 @@ class GfxInternal
                 fclose(fd); // write out PDB for shader debugging
             }
             // Delete older pdb files
+            static std::mutex shader_pdb_cleanup;
+            std::scoped_lock<std::mutex> lock(shader_pdb_cleanup);
```
gboisse, line 9885: That works, although any reason for not using `std::lock_guard` instead?

This is more out of curiosity, although it'd make the code slightly simpler, as `lock_guard` doesn't require the template argument iirc.

gfx.cpp
```diff
@@ -9936,18 +9943,18 @@ class GfxInternal
...
                     {
-                        std::sort(bytecode_files.begin(), bytecode_files.end(), [](const std::filesystem::path &a, const std::filesystem::path &b) {
-                            return last_write_time(a) > last_write_time(b); });
+                        std::sort(bytecode_files.begin(), bytecode_files.end(), [&ec](const std::filesystem::path &a, const std::filesystem::path &b) {
+                            return last_write_time(a, ec) > last_write_time(b, ec); });
                         for(size_t i = max_cached_files; i < bytecode_files.size(); ++i)
-                            std::filesystem::remove(bytecode_files[i]);
+                            std::filesystem::remove(bytecode_files[i], ec);
                     }
                     if (reflection_files.size() > max_cached_files)
```
gboisse: Remove space after `if`.

gboisse (approved):

gfx.cpp
maoliver-amd: No reason at all, scoped_lock is just the newer type and for locks like this I always just think 'scoped' with anything RAII.
As for the template args, lock_guard would need them aswell as the compiler was complaining for some reason about needing the type

gfx.cpp
maoliver-amd: wasnt originally part of this PR, but fixed anyway

PR #228, opened 2026-08-11
Fix auto-resize textures

gboisse:
Previous implementation was triggering the following debug layer error:
```
D3D12 ERROR: ID3D12CommandQueue1::ExecuteCommandLists: Placed resources, reserved resources, or committed resources with D3D12_HEAP_FLAG_CREATE_NOT_ZEROED flag with either render target or depth stencil flags must be initialized with a Discard/Clear/Copy operations before other operations are supported. Resource (0x0000022B6278C180:'Unnamed ID3D12Resource Object'), Subresource (0) is not initialized but is used in Function (AssertResourceState) on Command List (0x0000022948326B70:'gfx_CommandList'). [ EXECUTION ERROR #1422: RENDER_TARGET_OR_DEPTH_STENCIL_RESOUCE_NOT_INITIALIZED]
```

This was because `DiscardResource()` wasn't being called upon re-creating the "auto-resize" textures (for render targets and UAVs).

This PR moves part of the resizing logic prior to submitting the command list (so we can still schedule `DiscardResource()` commands, which isn't possible in the original resize callback as, at that point, the command list is already closed) and keeping the original callback, now renamed to `resizeBackBuffers()`.

#### Repro steps

Easiest way to repro is to add `kGfxCreateContextFlag_EnableDebugLayer` to either the `01-rtao` or `02-pbr` example, run the application, and resize the window.

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -4945,6 +4945,11 @@ public:
             {
                 RECT window_rect = {};
                 GetClientRect(window_, &window_rect);
+                uint32_t const window_width  = GFX_MAX(window_rect.right,  (LONG)8);
+                uint32_t const window_height = GFX_MAX(window_rect.bottom, (LONG)8);
+                bool const resized = (!IsIconic(window_) && (window_width != window_width_ || window_height != window_height_));
+                if(resized)
+                    GFX_TRY(resizeTextures(window_width, window_height));
                 D3D12_RESOURCE_BARRIER resource_barrier = {};
                 resource_barrier.Type = D3D12_RESOURCE_BARRIER_TYPE_TRANSITION;
                 resource_barrier.Transition.Subresource = D3D12_RESOURCE_BARRIER_ALL_SUBRESOURCES;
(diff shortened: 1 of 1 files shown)


gfx.cpp
```diff
@@ -10232,6 +10255,22 @@ class GfxInternal
...
+                    transition_state = D3D12_RESOURCE_STATE_RENDER_TARGET;
+                else if((resource_desc.Flags & D3D12_RESOURCE_FLAG_ALLOW_DEPTH_STENCIL) != 0)
+                    transition_state = D3D12_RESOURCE_STATE_DEPTH_WRITE;
+                else if((resource_desc.Flags & D3D12_RESOURCE_FLAG_ALLOW_UNORDERED_ACCESS) != 0)
+                    transition_state = D3D12_RESOURCE_STATE_UNORDERED_ACCESS;
+                if(transition_state != D3D12_RESOURCE_STATE_COMMON)
+                {
+                    if(transitionResource(texture, transition_state, kTransitionType_Implicit))
+                        submitPipelineBarriers();   // transition our resources if needed
+                    command_list_->DiscardResource(texture.resource_, nullptr);
```
gboisse, line 10271: Add missing call to `DiscardResource()`.

gfx.cpp
```diff
@@ -10202,10 +10203,32 @@ class GfxInternal
...
     {
         if(!IsWindow(window_)) return kGfxResult_NoError;   // can't resize past window tear down
         GFX_ASSERT(swap_chain_ != nullptr);
+        for(uint32_t i = 0; i < max_frames_in_flight_; ++i)
+        {
+            collect(back_buffers_[i]);
+            back_buffers_[i] = nullptr;
+            freeRTVDescriptor(back_buffer_rtvs_[i]);
+        }
+        sync(); // make sure the GPU is done with the previous swap chain before resizing
```
tengel-amd, line 10216: should we close the current command buffer first?

gboisse: At that point in the code, it is already closed (see `frame()` implementation).

tengel-amd (approved):

gfx.cpp
tengel-amd: fair enough, was wondering if this could be called from elsewhere 👍

gboisse: That'd be illegal indeed 🙂 but this is an internal function, completely inaccessible to the API user.

maoliver-amd (approved):

PR #231, opened 2026-08-27
Added support for GPU upload heap

SShikhali:
Added support for `D3D12_HEAP_TYPE_GPU_UPLOAD` (https://microsoft.github.io/DirectX-Specs/d3d/D3D12GPUUploadHeaps.html).

No caller side changes are required for this feature to work. ReBAR needs to be supported by the system (also probably needs to be enabled in BIOS).
I didn't notice any performance/memory improvements on my RTX 5080. It'd be nice to test it on an APU, but I don't have access to any.

### DIFF
--- a/gfx.cpp
+++ b/gfx.cpp
@@ -131,6 +131,7 @@ class GfxInternal
     D3D12MA::Allocation **back_buffer_allocations_ = nullptr;
     DXGI_FORMAT back_buffer_format_ = DXGI_FORMAT_R8G8B8A8_UNORM;
     DXGI_COLOR_SPACE_TYPE color_space_ = DXGI_COLOR_SPACE_RGB_FULL_G22_NONE_P709;
+    D3D12_HEAP_TYPE upload_heap_type_ = D3D12_HEAP_TYPE_UPLOAD;
     uint32_t *back_buffer_rtvs_ = nullptr;
     bool is_interop_ = false;
     uint32_t back_buffer_index_ = 0;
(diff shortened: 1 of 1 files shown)


gfx.cpp
```diff
@@ -1360,6 +1361,15 @@ class GfxInternal
         if(mesh_features.MeshShaderTier < D3D12_MESH_SHADER_TIER_1) { if(mesh_device_ != nullptr) mesh_device_->Release(); mesh_device_ = nullptr; }
         if(mesh_device_ == nullptr && mesh_command_list_ != nullptr) { mesh_command_list_->Release(); mesh_command_list_ = nullptr; }
 
+        D3D12_FEATURE_DATA_D3D12_OPTIONS16 gpu_upload_heap_features = {};
+        device_->CheckFeatureSupport(D3D12_FEATURE_D3D12_OPTIONS16, &gpu_upload_heap_features, sizeof(D3D12_FEATURE_DATA_D3D12_OPTIONS16));
+        bool const gpu_upload_heap_supported = gpu_upload_heap_features.GPUUploadHeapSupported;
+        GFX_PRINTLN("GPU upload heap is supported: %s", gpu_upload_heap_supported ? "true" : "false");
+        // Prefer the GPU Upload Heap when available. it's CPU-writable heap like a regular upload heap, but resides in VRAM instead of system memory.
+        // So the GPU can read it back at full bandwidth. https://microsoft.github.io/DirectX-Specs/d3d/D3D12GPUUploadHeaps.html
```
gboisse, line 1369: Do we know the impact on the performance of writing from CPU?

Arguably if this is meant for unified memory systems, it's probably pretty fast for both devices 🤔

SShikhali: I assume that now CPU pays the price for moving data across PCI-E? Since now it writes the data directly to the GPU, instead of GPU reading it from RAM over PCI-E🤔 But again, I don't see any perf diff on Bistro & Zero-Day scenes.

And for APUs this should probably run better.

maoliver-amd: Its unlikely well see any perf from this as currently we are just using it for upload. We never had a separate copy queue so theres no synch fences that GPU upload heaps avoid, so no savings from that.
In theory we can save on some copies but would require additional changes as we are still doing an extra copy from upload buffers into default vram buffers which are unneeded with ReBAR. This would require additional changes so we can expose an extra user visible type that uses rebar if available and avoids a second copy but falls back silently to upload+copy to vram. This would be most useful for push constant buffers that are uploaded each frame (or reasonably frequently) or APUs

gboisse: I wonder if this would be most useful for streaming techniques (as in streaming from disk directly into such a heap-allocated memory, think texture streaming for instance).

PR #232, opened 2026-08-27
[GLTF] Fix Skin Inverse Bind Matrix Offset

tengel-amd:


gfx_scene.cpp
```diff
@@ -2101,12 +2101,15 @@ class GfxSceneInternal
                 skin.joints_[j] = node_handles.at(gltf_skin.joints[j]);
             }
             skin.inverse_bind_matrices_.resize(gltf_skin.inverse_bind_matrices->count);
-            assert(gltf_skin.inverse_bind_matrices->count * sizeof(glm::mat4)
```
tengel-amd, line 2104: This was assuming that we must have 1 view for each skin. This is not mandatory afaik, and we should be able to have a view for all matrices, each skin having a slice of it for example.

maoliver-amd (approved):

